PR #19, opened 2022-07-08
Build cleanup

sobychacko:


build.gradle
```diff
@@ -28,36 +22,47 @@ plugins {
 
 apply plugin: 'io.spring.nohttp'
 
-//def gitPresent = new File('.git').exists()
+def gitPresent = new File('.git').exists()
 
-//if(gitPresent) {
-//	apply plugin: 'org.ajoberstar.grgit'
-//}
+if(gitPresent) {
```
onobc, line 27: **[TinyNit]** Spacing

build.gradle
```diff
@@ -28,36 +22,47 @@ plugins {
...
+	docResourcesVersion = '0.2.5'
+
+	javadocLinks = [
+			'https://docs.oracle.com/en/java/javase/17/docs/api/',
+			'https://docs.spring.io/spring-framework/docs/current/javadoc-api/'
+	] as String[]
+
+	if (gitPresent) {
+		modifiedFiles =
+				files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') || f.name.endsWith('.kt') }
```
onobc, line 48: Do we have any Kotlin files in here currently? If not, and we don't plan on adding anytime soon, I would be in favor of removing any Kotlin related pieces.

build.gradle
```diff
@@ -28,36 +22,47 @@ plugins {
...
+	assertjVersion = '3.22.0'
+	awaitilityVersion = '4.2.0'
 	googleJsr305Version = '3.0.2'
 	hamcrestVersion = '2.2'
-	hibernateValidationVersion = '6.2.3.Final'
-	jacksonBomVersion = '2.13.2.20220328'
+	hibernateValidationVersion = '7.0.4.Final'
+	jacksonBomVersion = '2.13.3'
 	jaywayJsonPathVersion = '2.6.0'
 	junit4Version = '4.13.2'
```
onobc, line 58: Why do we have junit4 in here?

build.gradle
```diff
@@ -28,36 +22,47 @@ plugins {
...
 	junitJupiterVersion = '5.8.2'
```
onobc, line 59: I think we can instead pull in the junit bom under dependency mgmt which would get rid of some of these version specified I believe. Same as what [spring-framework](https://github.com/spring-projects/spring-framework/blob/73d92d66b938a5ff9080442396bed13c421d1f8f/build.gradle#L309) does.

build.gradle
```diff
@@ -208,15 +210,68 @@ subprojects { subproject ->
...
+							it.appendNode('version', dep.version)
+							it.appendNode('scope', 'provided')
+						}
+					}
+				}
+			}
+		}
+	}
+
+	task updateCopyrights {
```
onobc, line 234: Nice addition!

build.gradle
```diff
@@ -296,3 +352,14 @@ project ('spring-pulsar-sample-apps') {
 	}
 }
 
+sonarqube {
```
onobc, line 355: Nice addition! I saw this and the copyrights in spring-integration and was wanting to add it here as well. You beat me to it. :)

spring-pulsar-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfiguration.java
```diff
@@ -24,15 +24,17 @@
...
 import org.springframework.pulsar.config.PulsarClientConfiguration;
 import org.springframework.pulsar.config.PulsarClientFactoryBean;
+import org.springframework.pulsar.core.DefaultPulsarConsumerFactory;
+import org.springframework.pulsar.core.DefaultPulsarProducerFactory;
 import org.springframework.pulsar.core.PulsarConsumerFactory;
 import org.springframework.pulsar.core.PulsarProducerFactory;
 import org.springframework.pulsar.core.PulsarTemplate;
 
 /**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration Auto-configuration} for Apache Pulsar.
```
onobc, line 36: **[Nit/Question]** I wonder if it is sufficient to just link to [org.springframework.boot.autoconfigure.Autoconfiguration](/Users/cbono/.gradle/caches/modules-2/files-2.1/org.springframework.boot/spring-boot-autoconfigure/3.0.0-SNAPSHOT/af0fa2b33c020b4e6b128c54eadcbce0dcb48f47/spring-boot-autoconfigure-3.0.0-SNAPSHOT-sources.jar!/org/springframework/boot/autoconfigure/AutoConfiguration.java:64) now that its a 1st class thing in SB3? 

Also, does this have to be fully specified like it is?

build.gradle
```diff
@@ -162,12 +167,9 @@ subprojects { subproject ->
 		testImplementation "org.hamcrest:hamcrest-core:$hamcrestVersion"
```
onobc, line 167: For the actual dependencies, should we specify the versions in dependencyManagement section and then just specify group/artifiact in the actual dependencies?

onobc (changes requested):
Nice cleanup @sobychacko . A few comments/questions but other than that, LGTM.

build.gradle
sobychacko: will remove.

build.gradle
sobychacko: will remove.

build.gradle
sobychacko: will address.

build.gradle
sobychacko: Ok, i just followed what Spring-Kafka does. We can do that. Feel free to make that change once this PR is merged.

spring-pulsar-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfiguration.java
sobychacko: Sure, we can do it that way. Feel free to do that as part of the cleanup that you will do in the boot module.

build.gradle
sobychacko: may not have to - i followed what SK does.

PR #25, opened 2022-07-11
Refactor template/producer to not use default topic name

onobc:
This proposal does the following (reason in sub-bullets):

* Introduces a PulsarOperations interface
  * matches spring-kafka and defines contract
  * provides default delegation methods to avoid cookie cutter code in impl
* Removes producer caching from PulsarTemplate (will add this back in on follow commit but probably in producer factory)
  * Current caching is danger of OOME (no cleaning of cache)
* Producer factory no longer holds onto a single producer instance
  * we hand out N of these, if anything we should hold reference to all and close etc..
* Removes defaultClientName from PulsarTemplate
  * not thread-safe (topic set on producer config in producer factory)
  * misleading API and it mutates the config props (I never liked this about KafkaTemplate)

### Fun facts 
* PulsarClient holds onto all Producers it spits out
* PulsarClient closes all producers on its close
* PulsarClient removes producer from list when producer is closed directly
* Pulsar producers are bound to a single topic at creation time

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
@@ -16,125 +16,88 @@
 
 package org.springframework.pulsar.core;
 
-import java.util.Map;
-import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
 
+import org.apache.commons.logging.LogFactory;
 import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageRouter;
 import org.apache.pulsar.client.api.Producer;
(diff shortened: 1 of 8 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -16,125 +16,88 @@
...
 	}
 
-	public void setSchema(Schema<T> schema) {
-		this.schema = schema;
+	private Producer<T> prepareProducerForSend(String topic, T message, MessageRouter messageRouter) throws PulsarClientException {
+		Schema<T> schema = SchemaUtils.getSchema(message);
+		return this.producerFactory.createProducer(topic, schema, messageRouter);
 	}
 
-	private class SchemaTopic {
```
onobc, line 101: We will bring this back when/if we bring back producer caching. I would be in favor of putting the caching in a CachingProducerFactory impl.

PR #38, opened 2022-07-24
Add test cases to PulsarTemplateTests

onobc:
This adds tests cases for the newly added `PulsarOperations` APIs.

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
@@ -38,7 +38,7 @@ public interface PulsarOperations<T> {
 	 * @throws PulsarClientException if an error occurs
 	 */
 	default MessageId send(T message) throws PulsarClientException {
-		return send(null, message);
+		return send(null, message, null);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
```diff
@@ -48,7 +48,30 @@ default MessageId send(T message) throws PulsarClientException {
...
+	}
+
+	/**
+	 * Sends a message to the default topic in a blocking manner.
+	 * @param message the message to send
+	 * @param messageRouter the optional message router to use
+	 * @return the id of the sent message
+	 * @throws PulsarClientException if an error occurs
+	 */
+	default MessageId send(T message, MessageRouter messageRouter) throws PulsarClientException {
```
onobc, line 62: I missed this variant on the original proposal.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
```diff
@@ -38,7 +38,7 @@
 	 * @throws PulsarClientException if an error occurs
 	 */
 	default MessageId send(T message) throws PulsarClientException {
-		return send(null, message);
+		return send(null, message, null);
```
onobc, line 41: I made all of the default impls call to the abstract one rather than daisy-chan into the next most specific `send` method. This prevents coupling between the different APIs (ie `send(T message)` only depends on the abstract `send(null, message, null);` whereas before it depended on `send(null, message);` which then depended on `send(null, message, null);` which then would mean that we have to consider all 3 when changing one of them. Basically 2 dependent function calls is less coupling than 3.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarProducerFactoryTests.java
```diff
@@ -45,12 +46,19 @@ abstract class PulsarProducerFactoryTests extends AbstractContainerBaseTests {
...
 	@BeforeEach
-	void setup() throws PulsarClientException {
+	void createPulsarClient() throws PulsarClientException {
 		pulsarClient = PulsarClient.builder()
 				.serviceUrl(getPulsarBrokerUrl())
 				.build();
 	}
 
+	@AfterEach
+	void closePulsarClient() throws PulsarClientException {
```
onobc, line 56: **[UNRELATED]** I noticed this test was not cleaning up when done w/ the client.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
```diff
@@ -17,110 +17,111 @@
...
+import org.junit.jupiter.api.Named;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 /**
+ * Tests for {@code PulsarTemplate}.
+ *
  * @author Soby Chacko
+ * @author Chris Bono
```
onobc, line 52: @soby we can go over this one via Zoom if you like. It basically removes the "testUsage" test and replaces the 2 sync/async cases w/ the permutations of `(syncOrAsync, defaultOrSpecificTopic, withOrWithoutRouter)` via `@ParameterizedTest`.

soby: @sobychacko

onobc: My bad @soby . Sorry for the interruption - have a nice day. 😸

sobychacko (approved):

sobychacko:
LGTM - Merged upstream.

PR #40, opened 2022-07-26
Initial reference docs attempt

sobychacko:


build.gradle
```diff
@@ -344,6 +344,17 @@ project ('spring-pulsar-boot-autoconfigure') {
 	}
 }
 
+project ('spring-pulsar-docs') {
+	description = 'Spring Pulsar Code Snippets For Docs'
+
+	dependencies {
+		api "org.springframework.boot:spring-boot-starter:$springBootVersion"
```
onobc, line 351: I wonder if we should instead use the spring-pulsar starter/auto-configure here to future proof additions from there?

publish-maven.gradle
```diff
@@ -0,0 +1,49 @@
+apply plugin: 'maven-publish'
+apply plugin: 'com.jfrog.artifactory'
+
+publishing {
```
onobc, line 4: We don't want this one as we already are doing this via https://github.com/spring-projects-experimental/spring-pulsar/blob/main/gradle/publish-artifactory.gradle

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/7e0c81ad1c1df71f946665718099ff4f5a2a3cec

PR #43, opened 2022-07-28
Use schema hash in producer cache key

onobc:
This proposal enhances the producer cache key as follows:
1. Renamed to `ProducerCacheKey`
2. Uses `SchemaHash` to take into account schema bytes and schema type

ℹ️ NOTE: @alpreu @sobychacko we recently talked about doing something more sophisticated for the `MessageRouter` as part of the cache key. I ended up dropping that because those impls we were looking at (single, roundrobin) are really internal routers and not meant to be set by calling `ProducerBuilder.messageRouter`. Rather, these routers are custom (aka user-provided). If the custom router does not implement its own `EQ/HC` then it will fallback to identity `EQ/HC`. In a Spring based world it is highly likely that the router is a bean and there is only a single instance and this would cache well. If the user is passing in different router instances then its possible we could get cache misses. I am ok w/ documenting this fact and moving forward for now. Thoughts?

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
@@ -18,11 +18,9 @@ package org.springframework.pulsar.core;
 
 import java.time.Duration;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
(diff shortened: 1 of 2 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
```diff
@@ -151,18 +152,50 @@ private void closeProducer(Producer<T> producer) {
...
+	 * Uniquely identifies a producer that was handed out by the factory.
 	 *
 	 * @param <T> type of the schema
 	 */
-	record SchemaTopic<T>(Schema<T> schema, String topicName, MessageRouter messageRouter) {
-		public SchemaTopic {
+	static class ProducerCacheKey<T> {
+
+		private final Schema<T> schema;
+		private final SchemaHash schemaHash;
```
onobc, line 162: Unfortunately, having the create a new instance field (`schemaHash`) is not supported in records. Had to go back to hand-written POJO for now.

alpreu:
The changes look good to me 👍 

As discussed offline, I asked around and `SchemaHash` is considered part of the private API. For the moment, it should be fine to rely on it anyway. We can go ahead with creating a PIP for introducing a `hashCode` for the Schema to the public API. This way we can continue development now and exchange the implementation detail later. I can create a draft for the proposal and send it over for you to review. What do you think?

onobc:
> The changes look good to me 👍
> 
> As discussed offline, I asked around and `SchemaHash` is considered part of the private API. For the moment, it should be fine to rely on it anyway. We can go ahead with creating a PIP for introducing a `hashCode` for the Schema to the public API. This way we can continue development now and exchange the implementation detail later. I can create a draft for the proposal and send it over for you to review. What do you think?

Thanks @alpreu , that sounds great.

sobychacko (approved):

sobychacko:
@alpreu Sounds like a plan. Please follow up with the PIP idea.

Issue #26, opened 2022-07-12
Allow users to produce messages with more options than simple payload


onobc:
The message API on Pulsar is rich. So we are currently allowing users to pass in payload. However there are many other options that we may want to expose to users. 
```java
TypedMessageBuilder<T> messageBuilder = producer.newMessage().value(message);
```
is what we in effect are using now. Suppose user wants to set `key` or other available properties on the message. There is no way for them to do that w/ PulsarTemplate currently. I will create an issue that outlines this.

I think we will want to allow user to specify a message customizer that will take the `TypedMessageBuilder` and let them do their thing when/if they want to. Something like this...

```java
Function<TypedMessageBuilder<T>, TypedMessageBuilder<T>> messageBuilderCustomizer = (mb) -> mb.key("foo");
TypedMessageBuilder<T> messageBuilder = producer.newMessage().value(message);
messageBuilder = messageBuilderCustomizer.apply(messageBuilder);
```

_Originally posted by @onobc in https://github.com/spring-projects-experimental/spring-pulsar/pull/25#discussion_r918306926_

PR #44, opened 2022-07-28
Implement keyed messages

alpreu:
This PR should implement keyed messaging on the PulsarTemplate (#26). 
A detailed description follows.

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
@@ -71,7 +71,20 @@ public interface PulsarOperations<T> {
 	 * @return the id of the sent message
 	 * @throws PulsarClientException if an error occurs
 	 */
-	MessageId send(String topic, T message, MessageRouter messageRouter) throws PulsarClientException;
+	default MessageId send(String topic, T message, MessageRouter messageRouter) throws PulsarClientException {
+		return send(topic, message, messageRouter, null);
+	}
+
+	/**
+	 * Sends a message to the specified topic in a blocking manner.
+	 * @param topic the topic to send the message to or {@code null} to send to the default topic
+	 * @param message the message to send
+	 * @param messageRouter the optional message router to use
+	 * @param typedMessageBuilderCustomizer the optional TypedMessageBuilder customizer
+	 * @return the id of the sent message
+	 * @throws PulsarClientException if an error occurs
+	 */
+	MessageId send(String topic, T message, MessageRouter messageRouter, TypedMessageBuilderCustomizer<T> typedMessageBuilderCustomizer) throws PulsarClientException;
 
 	/**
 	 * Sends a message to the default topic in a blocking manner.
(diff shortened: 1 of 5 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
```diff
@@ -71,7 +71,20 @@ default MessageId send(T message, MessageRouter messageRouter) throws PulsarClie
...
+	/**
+	 * Sends a message to the specified topic in a blocking manner.
+	 * @param topic the topic to send the message to or {@code null} to send to the default topic
+	 * @param message the message to send
+	 * @param messageRouter the optional message router to use
+	 * @param typedMessageBuilderCustomizer the optional TypedMessageBuilder customizer
+	 * @return the id of the sent message
+	 * @throws PulsarClientException if an error occurs
+	 */
+	MessageId send(String topic, T message, MessageRouter messageRouter, TypedMessageBuilderCustomizer<T> typedMessageBuilderCustomizer) throws PulsarClientException;
```
onobc, line 87: Wdyt about adding a few more combos to accommodate use of customizer w/o router (and vice-versa)?

**Current combos:**
```
- message
- message / topic
- message / router 
- message / topic / router
```
**New combos something like this:** 
``` 
- message / customizer
- message / topic / customizer
- message / customizer / router
- message / topic / customizer / router
```
The tradeoff of course is signature explosion vs. user convenience. For me, the concerns of signature explosion is lightened by the `default impls delegate to the single abstract w/ all of the params` pattern we are using.

The other tiny change I think we can make it put the customizer in front of the router in the arg list as I bet router is used less than customizer and the customizer would come directly after the payload arg (cohesive).

spring-pulsar/src/main/java/org/springframework/pulsar/core/TypedMessageBuilderCustomizer.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.core;
+
+import java.util.function.Function;
+import org.apache.pulsar.client.api.TypedMessageBuilder;
+
+/**
+ * The function to customize a {@link TypedMessageBuilder}.
```
onobc, line 23: I used an actual function in the backing issue, but that was mostly for illustration. I think we can use something similar to what Boot uses such as [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/DefaultKafkaProducerFactoryCustomizer.java). 

The 2 important things in there are 
* `@FunctionalInterface` marker on the interface
* Signature change slightly to `void customize(TheThingWeAreCustomizingHere<?, ?> theThing)`

onobc (changes requested):
That was fast @alpreu ! Thanks, looks great. I have a few minor comments/suggestions to follow up on. Additionally,

1. Please add yourself as `@author` on all modified files. 
2. Do you mind adding some tests cases to the existing PulsarTemplateTests?

Thanks again for the contribution!

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarOperations.java
alpreu: Thanks for the suggestions, and yes, having the customizer together with the payload makes sense 👍

spring-pulsar/src/main/java/org/springframework/pulsar/core/TypedMessageBuilderCustomizer.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.pulsar.client.api.TypedMessageBuilder;
+
+/**
+ * The interface to customize a {@link TypedMessageBuilder}.
+ *
+ * @author Alexander Preuß
+ */
+@FunctionalInterface
+public interface TypedMessageBuilderCustomizer {
```
onobc, line 27: Wdyt about adding the type `<T>` here and parameterize the parameter below w/ it?

onobc (changes requested):
Looks great @alpreu . Thanks for the updates and adding the tests. The only comment I have is around type info to the customizer. Once we resolve that we can remove the draft status and I will get it merged in.

alpreu:
Thank you for the feedback, I added the type to the message builder and updated the docs to reflect the changes. If it looks good to you I'll squash and rebase so you can merge :)

onobc (approved):

onobc:
Looks good @alpreu . I went ahead and squash/rebase/merge. Thanks again for contribution!

Issue #35, opened 2022-07-23
Provide the ability to set ProducerInterceptors on PulsarTemplate


sobychacko:
Consider adding the producer interceptor[s] on the Pulsar producer factory through `PulsarTemplate`.  When the producer is created in the factory, the interceptor[s] are passed on to the producer.

alpreu:
I'd like to have a go at this, you can assign the ticket to me :)

onobc:
There is the consumer [counterpart](https://github.com/spring-projects-experimental/spring-pulsar/issues/36) as well. I am not sure if there is any overlap etc.. but it would be good to keep that in mind when design/impl etc..

PR #59, opened 2022-08-10
Add ProducerInterceptors to PulsarTemplate and PulsarProducerFactory

alpreu:
This PR introduces the ability to add ProducerInterceptors to the underlying producer through the PulsarTemplate and PulsarFactory (#35).

### DIFF
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -142,8 +142,6 @@ You can disable the autoconfigured producer factory by providing your own bean d
 Producer<T> createProducer(String topic, Schema<T> schema) throws PulsarClientException;
 
 Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter) throws PulsarClientException;
-
-Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter, List<ProducerInterceptor> producerInterceptors) throws PulsarClientException;
 ----
 ====
 
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -531,9 +531,9 @@ public void listen(Foo foo) {
 
 On the producer side also, for the Java primitive types, the framework can infer the Schema, but for any other types, you need set that on the `PulsarTemmplate`.
 
-=== Intercepting messages
+==== Intercepting messages
 
-==== Intercept messages on the Producer
+===== Intercept messages on the Producer
 Adding a `ProducerInterceptor`  allows you to intercept and mutate messages received by the producer before being published to the brokers.
 To do so, you can pass a list of interceptors into the `PulsarTemplate` constructor.
 When using multiple interceptors, the order they are applied in will be the order they appear in the list.
@@ -557,7 +557,7 @@ ProducerInterceptor secondInterceptor() {
 ```
 
 
-=== Appendix
+==== Appendix
 The reference documentation has the following appendices:
 
 [horizontal]
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -142,6 +142,8 @@ You can disable the autoconfigured producer factory by providing your own bean d
 Producer<T> createProducer(String topic, Schema<T> schema) throws PulsarClientException;
 
 Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter) throws PulsarClientException;
+
+Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter, List<ProducerInterceptor> producerInterceptors) throws PulsarClientException;
 ----
 ====
 
@@ -616,6 +618,32 @@ public void listen(Foo foo) {
 ```
 
 On the producer side also, for the Java primitive types, the framework can infer the Schema, but for any other types, you need set that on the `PulsarTemmplate`.
+
+=== Intercepting messages
+
+==== Intercept messages on the Producer
+Adding a `ProducerInterceptor`  allows you to intercept and mutate messages received by the producer before being published to the brokers.
+To do so, you can pass a list of interceptors into the `PulsarTemplate` constructor.
+When using multiple interceptors, the order they are applied in will be the order they appear in the list.
+
+If you are using Spring Boot auto-configuration, you can simply specify the interceptors as Beans.
+They will be passed automatically to the `PulsarTemplate`.
+Ordering of the interceptors is achieved by using the `@Order` annotation as seen below.
+
+```
+@Bean
+@Order(100)
+ProducerInterceptor firstInterceptor() {
+  ...
+}
+
+@Bean
+@Order(200)
+ProducerInterceptor secondInterceptor() {
+  ...
+}
+```
+
 
 === Appendix
 The reference documentation has the following appendices:
--- a/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
@@ -49,6 +49,7 @@ import org.springframework.pulsar.listener.DefaultPulsarMessageListenerContainer
  * Autoconfiguration tests for {@link PulsarAutoConfiguration}.
  *
  * @author Chris Bono
+ * @author Alexander Preuß
  */
 @SuppressWarnings("unchecked")
 class PulsarAutoConfigurationTests {
--- a/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
@@ -17,10 +17,15 @@
 package org.springframework.pulsar.autoconfigure;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.in;
 import static org.mockito.Mockito.mock;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.pulsar.client.api.interceptor.ProducerInterceptor;
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
@@ -152,6 +157,17 @@ class PulsarAutoConfigurationTests {
 				.run((context) -> assertThat(context).hasNotFailed()
 						.getBean(PulsarListenerAnnotationBeanPostProcessor.class)
 						.isSameAs(listenerAnnotationBeanPostProcessor));
+	}
+
+	@Test
+	void customProducerInterceptorIsUsedInPulsarTemplate() {
+		ProducerInterceptor interceptor = mock(ProducerInterceptor.class);
+		this.contextRunner
+				.withBean("customProducerInterceptor", ProducerInterceptor.class, () -> interceptor)
+				.run((context -> assertThat(context).hasNotFailed()
+						.getBean(PulsarTemplate.class)
+						.extracting("interceptors").asInstanceOf(InstanceOfAssertFactories.list(ProducerInterceptor.class))
+						.contains(interceptor)));
 	}
 
 	@Nested
--- a/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
@@ -17,11 +17,8 @@
 package org.springframework.pulsar.autoconfigure;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.in;
 import static org.mockito.Mockito.mock;
 
-import java.util.Collections;
-import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.pulsar.client.api.interceptor.ProducerInterceptor;
@@ -162,11 +159,10 @@ class PulsarAutoConfigurationTests {
 	@Test
 	void customProducerInterceptorIsUsedInPulsarTemplate() {
 		ProducerInterceptor interceptor = mock(ProducerInterceptor.class);
-		this.contextRunner
-				.withBean("customProducerInterceptor", ProducerInterceptor.class, () -> interceptor)
-				.run((context -> assertThat(context).hasNotFailed()
-						.getBean(PulsarTemplate.class)
-						.extracting("interceptors").asInstanceOf(InstanceOfAssertFactories.list(ProducerInterceptor.class))
+		this.contextRunner.withBean("customProducerInterceptor", ProducerInterceptor.class, () -> interceptor)
+				.run((context -> assertThat(context).hasNotFailed().getBean(PulsarTemplate.class)
+						.extracting("interceptors")
+						.asInstanceOf(InstanceOfAssertFactories.list(ProducerInterceptor.class))
 						.contains(interceptor)));
 	}
 
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
@@ -95,7 +95,8 @@ public class CachingPulsarProducerFactory<T> extends DefaultPulsarProducerFactor
 	public Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter,
 			List<ProducerInterceptor> producerInterceptors) {
 		final String topicName = ProducerUtils.resolveTopicName(topic, this);
-		ProducerCacheKey<T> producerCacheKey = new ProducerCacheKey<>(schema, topicName, messageRouter);
+		ProducerCacheKey<T> producerCacheKey = new ProducerCacheKey<>(schema, topicName, messageRouter,
+				producerInterceptors);
 		return this.producerCache.get(producerCacheKey, (st) -> {
 			try {
 				return this.doCreateProducer(st.topic, st.schema, st.router, producerInterceptors);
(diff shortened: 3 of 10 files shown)


alpreu:
There's a few things I'm unsure about, I'm not sure if the way I implemented it now is the 'Spring-way' because I was previously only using the framework, not developing it. Another thing to think about is if the interceptors should be added to the cache key as well. I'm not sure if this makes sense from a user perspective though (is there a reason to have 2 templates writing to the same topic but one with interceptors set and the other one without?).

Also still missing documentation, I will add it after we finalize the code review.

sobychacko:
@alpreu Thanks for the PR!! The approach looks good overall. Here are some comments I have.  cc @onobc 
Since producer interceptors are optional, what do you think about providing a mutator on `PulsarTemplate` rather than setting it through the constructor? I say that because if we go with the constructor approach, we may end up with more constructors for each optional concern like this in the future. But, let's discuss; I am open either way. 

Rather than taking a collection of `ProducerInterceptor` on `PulsarTemplate,` how about providing a setter for a single interceptor? I think that is probably the majority of the use case. For multiple interceptors on the same template, how about introducing another indirection - something like a `CompositeProducerInterceptor` that implements the `ProducerInterceptor`? In this `CompositeProduceInterceptor,` we can set the delegate interceptors in the order the user wants and then call each interceptor. 

For caching, I think it might be better to have less number of parameters to use for keys. If we have the same topic and schema/router used, that becomes the lowest denominator for the cache key, right? In that case, I think we can simply document that the same interceptors are used if the template is used in multiple contexts. We can relax this rule if this becomes a broader use case. 

Anyways, let's discuss and come to a consensus.

onobc:
@alpreu I am about to review the PR and noticed [the checkstyle violations](https://github.com/spring-projects-experimental/spring-pulsar/runs/7767008296?check_suite_focus=true#step:6:266). I am not writing this to balk about that but rather it made me realize that I did not communicate the fact that I added in Spring JavaFormat to the project when I did that build/gradle work last week. I am going to add it to a "Working with the code" wiki page soon. But in the meantime, the gist is this...

We are using [Spring JavaFormat](https://github.com/spring-io/spring-javaformat) for checkstyle and formatting. 

When you see errors like this:

```java
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':spring-pulsar:checkFormatMain'.
> Formatting violations found in the following files:
   * src/main/java/org/springframework/pulsar/core/PulsarProducerFactory.java
   * src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
   * src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
   * src/main/java/org/springframework/pulsar/core/DefaultPulsarProducerFactory.java
  
  Run `format` to fix.
```
You can then run the "format" task and it will typically auto-correct all of the violations. I have found this VERY NICE. From time-time it will not auto-fix all of them and you will have to do it manually. 

You can give it a test run on your branch when you get back to it and see it in action. I will get that doc created in the next 1-2 days as well.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -42,13 +44,25 @@
 
 	private final PulsarProducerFactory<T> producerFactory;
 
+	private final List<ProducerInterceptor> interceptors;
+
 	/**
```
onobc, line 49: > I'm not sure if the way I implemented it now is the 'Spring-way' because I was previously only using the framework, not developing it.

The one change I would make, is add the injection of these interceptors into the PulsarAutoConfiguration as such:
```java
@Bean
@ConditionalOnMissingBean(PulsarTemplate.class)
public PulsarTemplate<?> pulsarTemplate(PulsarProducerFactory<?> pulsarProducerFactory, ObjectProvider<ProducerInterceptor> interceptors) {
  return new PulsarTemplate<>(pulsarProducerFactory, interceptors.orderedStream().toList());
}
```
You will see this pattern prevalent in the Framework and especially in Boot when it comes to injecting one-or-more optional beans. The beauty of it is that it provides a nice stream API for lazy retrieving the beans. 
As you can see, the `interceptors.orderedStream().toList());` above will give an ordered list of any PI beans in the app context. 

Consider a user defining these interceptors beans:

```java
@Bean
@Order(200)
ProducerInterceptor interceptorFoo() {
  return null;
}

@Bean
@Order(100)
ProducerInterceptor interceptorBar() {
  return null;
}
```

then the list would contain `interceptorBar, interceptorFoo` as it respects the bean order.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -42,13 +44,25 @@
...
 	public PulsarTemplate(PulsarProducerFactory<T> producerFactory) {
+		this(producerFactory, null);
+	}
+
+	/**
+	 * Constructs a template instance.
+	 * @param producerFactory the producer factory used to create the backing Pulsar producers.
+	 * @param interceptors the {@link ProducerInterceptor}s to add to the producer.
+	 */
+	public PulsarTemplate(PulsarProducerFactory<T> producerFactory, List<ProducerInterceptor> interceptors) {
```
onobc, line 63: > Rather than taking a collection of ProducerInterceptor on PulsarTemplate, how about providing a setter for a single interceptor? I think that is probably the majority of the use case. For multiple interceptors on the same template, how about introducing another indirection - something like a CompositeProducerInterceptor that implements the ProducerInterceptor? In this CompositeProduceInterceptor, we can set the delegate interceptors in the order the user wants and then call each interceptor.

@sobychacko I do understand keeping it simple and not adding things we may not need. In this case, because of the natural fit w/ "N things" that the [`ObjectProvider`](https://github.com/spring-projects-experimental/spring-pulsar/pull/59#discussion_r942860574) provides as well as the fact that the underlying Pulsar API accept "N things" I would be in favor of keeping "N" from the start.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -42,13 +44,25 @@
...
 	/**
 	 * Constructs a template instance.
 	 * @param producerFactory the producer factory used to create the backing Pulsar
 	 * producers.
 	 */
 	public PulsarTemplate(PulsarProducerFactory<T> producerFactory) {
+		this(producerFactory, null);
+	}
+
+	/**
```
onobc, line 58: > Since producer interceptors are optional, what do you think about providing a mutator on PulsarTemplate rather than setting it through the constructor? I say that because if we go with the constructor approach, we may end up with more constructors for each optional concern like this in the future. But, let's discuss; I am open either way.

@sobychacko I understand wanting to avoid `<init>` bloat as it can get out of hand quick. In this case I am favor of keeping it as-is because 
1. it is a natural fit w/ the use of [`ObjectProvider`](https://github.com/spring-projects-experimental/spring-pulsar/pull/59#discussion_r942860574) in the auto-config
2. Similar to constructor bloat is mix of constructor and `setXYZ` (aka setter bloat). I think the proper way to avoid both constructor and setter bloat would be to introduce a builder pattern. However, I think we do not need to worry about that right now. 

Wdyt?

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
sobychacko: Ok, that makes sense. `ObjectProvider` usgae is nice.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
sobychacko: Yep, let's keep it that way. We will see how the template evolves.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarProducerFactory.java
```diff
@@ -60,10 +63,16 @@ public Producer<T> createProducer(String topic, Schema<T> schema) throws PulsarC
 	@Override
 	public Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter)
 			throws PulsarClientException {
-		return doCreateProducer(topic, schema, messageRouter);
+		return createProducer(topic, schema, messageRouter, null);
```
onobc, line 66: **[UNRELATED_NIT]** So this is something I introduced and did catch until now... 

When I use the `"N 'action' methods delegate to the 'doAction' impl method"` pattern (I made that term up just now 😉 ) I typically try to avoid the N level method chaining and just have each call to the `doAction` method directly. The main reason is it keeps the top-level "action methods" API decoupled a bit and makes it easier to evolve one of the apis w/o worrying about breaking the other API as well (aka smaller blast radius).

As it is currently, any change to `createProducer(String topic, Schema<T> schema, MessageRouter messageRouter, List<ProducerInterceptor> producerInterceptors)` would be considered a possible change to `createProducer(String topic, Schema<T> schema, MessageRouter messageRouter)` which in turn would bring `createProducer(String topic, Schema<T> schema)` into the potential blast radius. 

Albeit, the current impl has no logic so it would not be noticeable. However, if we start varying them by adding validation in each method then it could bite us later. 

ℹ️ Also note that we are doing the same in PulsarOperations/Template (just noticed this).

alpreu: This is a good point, I will also create a separate PR to change the PulsarOperations & -Template in this regard

spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
```diff
@@ -89,12 +92,12 @@ public CachingPulsarProducerFactory(PulsarClient pulsarClient, Map<String, Objec
 	}
 
 	@Override
-	public Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter) {
+	public Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter, List<ProducerInterceptor> producerInterceptors) {
```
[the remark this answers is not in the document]
alpreu: That's what I also fear, I'll go ahead and add it

alpreu:
Thank you for the review and also the pointers on Spring concepts, much appreciated!!

Regarding the `"N 'action' methods delegate to the 'doAction' impl method"` pattern, we do need to return `createProducer` with the full signature in the convenience methods. If we don't, the `CachingPulsarProducerFactory`s implementation that adds the producer to the cache does not get called.

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -617,6 +619,32 @@ public void listen(Foo foo) {
 
 On the producer side also, for the Java primitive types, the framework can infer the Schema, but for any other types, you need set that on the `PulsarTemmplate`.
 
+=== Intercepting messages
+
+==== Intercept messages on the Producer
```
onobc, line 624: Nice doc! Thanks for adding.

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -142,6 +142,8 @@ You can disable the autoconfigured producer factory by providing your own bean d
 Producer<T> createProducer(String topic, Schema<T> schema) throws PulsarClientException;
 
 Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter) throws PulsarClientException;
+
+Producer<T> createProducer(String topic, Schema<T> schema, MessageRouter messageRouter, List<ProducerInterceptor> producerInterceptors) throws PulsarClientException;
```
onobc, line 146: The latest docs commit removes many of the hardcoded areas where we specifying the APIs like in PPF. You can simply remove this line and the merge should take care of the rest (this section has been removed).

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarProducerFactory.java
```diff
@@ -75,6 +84,9 @@ protected Producer<T> doCreateProducer(String topic, Schema<T> schema, MessageRo
 		if (messageRouter != null) {
 			producerBuilder.messageRouter(messageRouter);
 		}
+		if (producerInterceptors != null && !producerInterceptors.isEmpty()) {
```
onobc, line 87: If you have not seen it yet, the `org.springframework.util.CollectionUtils` provides a convenience `isEmpty` that does both null/empty for us.

spring-pulsar/src/test/java/org/springframework/pulsar/core/CachingPulsarProducerFactoryTests.java
```diff
@@ -111,41 +113,68 @@ void createProducerWithMatrixOfCacheKeys() throws PulsarClientException {
 		Schema<String> schema2 = new StringSchema();
 		MessageRouter router1 = mock(MessageRouter.class);
 		MessageRouter router2 = mock(MessageRouter.class);
+		List<ProducerInterceptor> interceptors1 = List.of(mock(ProducerInterceptor.class));
+		List<ProducerInterceptor> interceptors2 = List.of(mock(ProducerInterceptor.class));
 
 		PulsarProducerFactory<String> producerFactory = producerFactory(pulsarClient, Collections.emptyMap());
 
-		// ask for the same 9 unique combos 3x - should end up w/ only 9 entries in cache
+		// ask for the same 21 unique combos 3x - should end up w/ only 21 entries in
```
onobc, line 121: Ugghhh... permutations 🤣 

First it's 3. Then it's 9. Then it's 21. Then it's...

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarProducerFactoryTests.java
```diff
@@ -120,4 +145,29 @@ protected void assertProducerHasTopicSchemaAndRouter(Producer<String> producer,
 	protected abstract PulsarProducerFactory<String> producerFactory(PulsarClient pulsarClient,
 			Map<String, Object> producerConfig);
 
+	@SuppressWarnings("rawtypes")
+	private static class MockInterceptor implements ProducerInterceptor {
```
onobc, line 149: Any reason to not just use `Mockito.mock`?

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
```diff
@@ -154,6 +157,16 @@ void customPulsarListenerAnnotationBeanPostProcessorIsRespected() {
 						.isSameAs(listenerAnnotationBeanPostProcessor));
 	}
 
+	@Test
+	void customProducerInterceptorIsUsedInPulsarTemplate() {
```
onobc, line 161: This looks great @alpreu . It may be helpful to add another variant that configures in 2 interceptors for sanity check. There is a variant of `withBean` that takes a `BeanDefinitionCustomizer` but that does not let you set the `@Order` for the bean. However, you can still use the `ApplicationContextRunner` approach w/ a slight twist like:
```java
	@Test
	void someTest() {
		this.contextRunner
				.withUserConfiguration(InterceptorTestConfiguration.class)
				.run((context) -> assertThat(context).hasNotFailed().hasBean("interceptorOne"));
	}

	@Configuration(proxyBeanMethods = false)
	static class InterceptorTestConfiguration {

		@Bean
		@Order(100)
		String interceptorOne() {
			return ...
		}
                 ....
	}
```
You may already know about this though. 

Also, this is not critical and if you don't have time to get to it I would be happy to add it during merge etc.. Just lmk.

onobc (approved):
@alpreu  I have a few minor comments but other than that, looks great. Thank you for the thoroughness (updating docs, updating/adding tests, etc..) it is noticed and much appreciated. 

> Thank you for the review and also the pointers on Spring concepts, much appreciated!!

You are more than welcome! 

> Regarding the "N 'action' methods delegate to the 'doAction' impl method" pattern, we do need to return createProducer with the full signature in the convenience methods. If we don't, the CachingPulsarProducerFactorys implementation that adds the producer to the cache does not get called.

Ah yeh, fair enough. This warrants bypassing that "default pattern". Thanks for looking into it.

sobychacko (approved):

spring-pulsar/src/test/java/org/springframework/pulsar/core/CachingPulsarProducerFactoryTests.java
alpreu: Yeah I didn't like that part either but I wanted to cover the interceptor cases

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarProducerFactoryTests.java
alpreu: I added this before I saw that we are using Mockito in other places. I'll swap it out

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
alpreu: How would you retrieve the bean instances from the configuration to confirm the order in the list? Or would you just check that there are two interceptors set in the Template? If you can add this part I'd be interested in taking a look at the solution :)

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/e32ed00750611ad5ae6e3aa8fa28a8b01dadfd3b

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
onobc: I added the test [here](https://github.com/spring-projects-experimental/spring-pulsar/commit/664c62eb3672ea08c080f2a636537dcb8c224e89#diff-89ca4830da9a2117dd3d4c118fffa9f7c67e601248e9aa7c5a231568855d998eR226). I used a static instance var to identify the interceptors. I know, accessing the statics like this does feel gross. On the flip side, the alternative that I know of is to make override some identifying method on the producer (see below for `toString`) also feels gross. Of the two choices I went w/ the former mostly because it is less code and a bit easier to follow than some magic toString override. 
```java


@Test
void customProducerInterceptorsOrderedProperly() {
  this.contextRunner
      .withUserConfiguration(InterceptorTestConfiguration.class)
      .run((context -> assertThat(context).hasNotFailed().getBean(PulsarTemplate.class)
          .extracting("interceptors")
          .asInstanceOf(InstanceOfAssertFactories.list(ProducerInterceptor.class))
          .extracting(ProducerInterceptor::toString)
          .containsExactly("ProducerInterceptor(Bar)", "ProducerInterceptor(Foo)")));
}

@Configuration(proxyBeanMethods = false)
static class InterceptorTestConfiguration {
  @Bean
  @Order(200)
  ProducerInterceptor interceptorFoo() {
    ProducerInterceptor interceptor = mock(ProducerInterceptor.class);
    when(interceptor.toString()).thenReturn("ProducerInterceptor(Foo)");
    return interceptor;
  }

  @Bean
  @Order(100)
  ProducerInterceptor interceptorBar() {
    ProducerInterceptor interceptor = mock(ProducerInterceptor.class);
    when(interceptor.toString()).thenReturn("ProducerInterceptor(Bar)");
    return interceptor;
  }
}
```

alpreu: Ah I see. Yeah, the toString-way is a bit hacky I agree

Issue #41, opened 2022-07-27
Provide NewTopic functionality


onobc:
Similar to Spring Kafka, we should provide ability to auto-create topics for any `@NewTopic` beans configured by the user.

PR #64, opened 2022-08-15
Auto-create Pulsar topics

alpreu:
WIP PR for #41 , not review ready


sobychacko:
Thanks, @alpreu. This is a great start. This is precisely something we had in mind as a high-priority feature item. Much appreciated you started with it. Please let us know once it is ready for review.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationTests.java
```diff
@@ -0,0 +1,116 @@
...
+
+	List<String> getTopics() throws Exception {
+		try (PulsarAdmin admin = PulsarAdmin.builder().serviceHttpUrl(getHttpServiceUrl()).build()) {
+			return admin.topics().getList(NAMESPACE);
+		}
+	}
+
+	void assertTopicsExist(List<String> actualTopics, ObjectProvider<PulsarTopic> expected) {
+		List<String> expectedTopics= expected.stream()
+				.<String>mapMulti((topic, consumer) -> {
```
onobc, line 44: Ah nice! Clever use of `mapMulti`. **TIL**.

What you have done w/ this test is an excellent application of the nested test stuff we were chatting about offline. I took the liberty to take what you have here and tweak it a bit to come w/ something that simplifies it a bit. I will post the code here in its entirety rather than try to comment line by line by line. 

```java
package org.springframework.pulsar.core;

import static org.assertj.core.api.Assertions.assertThat;
import java.util.List;
import java.util.Map;
import org.apache.pulsar.client.admin.PulsarAdmin;
import org.apache.pulsar.client.admin.PulsarAdminException;
import org.apache.pulsar.client.api.PulsarClientException;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Nested;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.springframework.beans.factory.ObjectProvider;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.context.ApplicationContext;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.test.context.ContextConfiguration;
import org.springframework.test.context.junit.jupiter.SpringExtension;


@ExtendWith(SpringExtension.class)
@ContextConfiguration
public class PulsarAdministrationTests extends AbstractContainerBaseTests {

	private static final String NAMESPACE = "public/default";

	@Autowired
	private PulsarAdmin pulsarAdminClient;

	@Configuration(proxyBeanMethods = false)
	static class AdminConfiguration {
		@Bean
		PulsarAdmin pulsarAdminClient() throws PulsarClientException {
			return PulsarAdmin.builder().serviceHttpUrl(getHttpServiceUrl()).build();
		}
	}

	private void assertThatTopicsExist(List<PulsarTopic> expected) throws PulsarAdminException {
		List<String> expectedTopics= expected.stream()
				.<String>mapMulti((topic, consumer) -> {
					if (topic.isPartitioned()) {
						for (int i = 0; i < topic.numberOfPartitions(); i++) {
							consumer.accept(topic.getFullyQualifiedTopicName() + "-partition-" + i);
						}
					} else {
						consumer.accept(topic.getFullyQualifiedTopicName());
					}

				}).toList();
		assertThat(pulsarAdminClient.topics().getList(NAMESPACE)).containsAll(expectedTopics);
	}

	@Nested
	@ContextConfiguration(classes = CreateMissingTopicsTest.CreateMissingTopicsConfig.class)
	class CreateMissingTopicsTest {

		@Test
		void testTopics(@Autowired ObjectProvider<PulsarTopic> expectedTopics) throws Exception {
			assertThatTopicsExist(expectedTopics.stream().toList());
		}

		@Configuration(proxyBeanMethods = false)
		static class CreateMissingTopicsConfig {

			@Bean
			PulsarTopic nonPartitionedTopic() {
				return PulsarTopic.builder("cmt-non-partitioned-1").build();
			}

			@Bean
			PulsarTopic nonPartitionedTopic2() {
				return PulsarTopic.builder("cmt-non-partitioned-2").build();
			}

			@Bean
			PulsarTopic partitionedTopic() {
				return PulsarTopic.builder("cmt-partitioned-1").setNumberOfPartitions(4).build();
			}
		}

	}

	@Nested
	@ContextConfiguration(classes = IncrementPartitionCountTest.IncrementPartitionCountConfig.class)
	class IncrementPartitionCountTest {

		@BeforeEach
		void createSmallerTopic() {
			try {
				pulsarAdminClient.topics().createPartitionedTopic("ipc-partitioned-1", 2);
			} catch (Exception e) {
				throw new RuntimeException(e);
			}
		}
		
		@Test
		void testTopics(@Autowired ObjectProvider<PulsarTopic> expectedTopics) throws Exception {
			assertThatTopicsExist(expectedTopics.stream().toList());
		}

		@Configuration(proxyBeanMethods = false)
		static class IncrementPartitionCountConfig {
			@Bean
			PulsarTopic partitionedTopic() {
				return PulsarTopic.builder("ipc-partitioned-1").setNumberOfPartitions(4).build();
			}
		}
	}
}

```

As you can see, by lifting up the PulsarAdmin into an actual bean and self-injecting at the top-level we can just use that in our assert method now. Also, the static was replaced w/ a BeforeEach. 

Let me know what you think.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationBadContextTests.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+
+public class PulsarAdministrationBadContextTests extends AbstractContainerBaseTests {
+
+		@Test
+		void testDecrementingPartitionCount() {
```
onobc, line 15: I see why you pulled this out to its own test as its a bit diff model and it kills the context during load. I think we can pull this back into the other test by doing the following restructuring:

1. Instead of pre-create a large one and then try to create a smaller one at load time we can create a large one during load time then in the test method call PulsarAdministration.createOrModifyTopics w/ the smaller topic and assert failure at that point. Something like this: 
```java
	@Nested
	@ContextConfiguration(classes = DecrementPartitionCount.DecrementPartitionCountConfig.class)
	class DecrementPartitionCount {

		@Test
		void testDecrementingPartitionCount(@Autowired ObjectProvider<PulsarTopic> expectedTopics) throws Exception {
			assertThatTopicsExist(expectedTopics.stream().toList());
			PulsarTopic smallerTopic = PulsarTopic.builder("dpc-partitioned-1").setNumberOfPartitions(4).build();
			assertThatIllegalStateException()
					.isThrownBy(() -> pulsarAdministration.createOrModifyTopics(smallerTopic))
					.withMessage("Topic persistent://public/default/dpc-partitioned-1 found with 8 partitions. Needs to be deleted first.");
		}

		@Configuration(proxyBeanMethods = false)
		static class DecrementPartitionCountConfig {
			@Bean
			PulsarTopic partitionedTopic() {
				return PulsarTopic.builder("dpc-partitioned-1").setNumberOfPartitions(8).build();
			}
		}
	}
```
wdyt?

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTopicTests.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.stream.Stream;
+import org.apache.pulsar.common.naming.TopicDomain;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class PulsarTopicTests {
+
+	@Test
```
onobc, line 13: What happens when you add a test w/ a string name of "some-topic-name/" (that ends in a slash)?

onobc (changes requested):
@alpreu awesome work! I like how you applied the nested testconfig. I made some simplification requests on them as well. This is looking good though. Ping me if you have any questions and thank you for this contribution.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -0,0 +1,171 @@
...
+		this.adminBuilder = PulsarAdmin.builder().loadConf(adminConfig);
+	}
+
+	public PulsarAdministration(PulsarAdminBuilder adminBuilder) {
+		this.adminBuilder = adminBuilder;
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		initialize();
```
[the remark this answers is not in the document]
onobc: On 2nd thought after seeing the tests - we can live w/o this for now.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -0,0 +1,171 @@
...
+import org.apache.pulsar.client.admin.PulsarAdminException;
+import org.apache.pulsar.client.api.PulsarClientException;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.core.log.LogAccessor;
+import org.springframework.util.CollectionUtils;
+
+public class PulsarAdministration implements ApplicationContextAware, SmartInitializingSingleton, PulsarAdminOperations {
```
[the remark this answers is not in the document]
onobc: To follow on - I realized that the pulsar admin client is named `PulsarAdmin` so you chose this path to avoid a duplilcate name only differentiated by packages - which I thank you for. having to fully qualify in the code is not a pretty thing. I am ok w/ this name now. Please close this thread once you digest it.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		PulsarAdmin admin = null;
+		try {
+			admin = createAdminClient();
+		} catch (Exception e) {
+			throw new IllegalStateException("Could not create PulsarAdmin", e);
+		}
+
+		if (admin != null) {
```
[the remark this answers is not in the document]
alpreu: Good catch, thanks

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationTests.java
alpreu: Thanks Chris, this definitely looks better than my static setup and makes the test easier to understand 👍

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationBadContextTests.java
alpreu: Yes, initially I wanted to have all of them in the same place but failed to do so with the way I had the test structured. It's great that we can have it all in one place now!

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -0,0 +1,171 @@
...
+							} else if (numberOfExistingPartitions > topic.numberOfPartitions()) {
+								throw new IllegalStateException(
+										"Topic " + topic.getFullyQualifiedTopicName() + " found with "
+												+ numberOfExistingPartitions
+												+ " partitions. Needs to be deleted first.");
+							}
+						}
+					} else {
+						if (!existingTopicsInNamespace.contains(topic.getFullyQualifiedTopicName())) {
+							logger.info("Topic " + topic.getFullyQualifiedTopicName() + " does not exist.");
```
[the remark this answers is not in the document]
onobc: Ahh yeh, not enough detail in my initial comment. 

So, prior to log accessor, if I wanted to conditionally debug log to avoid possible expensive log message creation I would have to pollute the code w/ `if logger.isXYZEnabled()` which made the code terse and I would often not use it in favor of more readable code. 
```
if (logger.isDebugEnabled()) {
	logger.debug("expensive string goes here");
}
```

However, the beauty of LogAccessor is this API to pass in supplier and under the hood it does the `if logger.isXYZEnabled()` and solves both clean code and better performing tradeoff of the previous logger APIs. 
```java
logger.debug(() -> "expensive string goes here");
```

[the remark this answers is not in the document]
onobc: By "lambda" I really meant "the log accessor API that takes a supplier."

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -0,0 +1,171 @@
...
+			return;
+		}
+
+		Map<String, List<PulsarTopic>> topicsPerNamespace = getTopicsPerNamespace(topics);
+
+		Set<PulsarTopic> topicsToCreate = new HashSet<>();
+		Set<PulsarTopic> topicsToModify = new HashSet<>();
+
+		topicsPerNamespace.forEach((namespace, requestedTopics) -> {
+			try (admin) {
```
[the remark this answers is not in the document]
alpreu: Fully agree here, I missed the autocloseable

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTopicTests.java
onobc: Yeh this crossed my mind as well. One tradeoff on relying on the formatting in `toString` for some business logic is that it is usually not clear that is the case. We can always add docs in javadocs but it is not as direct as having a separate API. In this case I would have to see the code change to make a decision. I would say go w/ whatever feels best and we can chat more about it then.

sobychacko:
@alpreu, I know this PR is WIP, but some general comments. Please add copyrights and author tags to all the classes. Also, add javadocs on public API methods/classes, etc. We should also make the `PulsarAdministration` bean auto-configured.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -0,0 +1,156 @@
...
+					}
+				}
+
+				createTopics(admin, topicsToCreate);
+				modifyTopics(admin, topicsToModify);
+			} catch (PulsarAdminException e) {
+				throw new RuntimeException(e);
+			}
+		});
+		admin.close();
```
onobc, line 132: Not the end of the world probably, but this is will not be closed when topic creation fails above. The downside is that if we put this whole block in a giant try/resources or try/finally is that it way indented and less readable. One technique I do here is to move the logic to a `doXYZ` and then wrap it with a try/resources such as: 

```java
try (PulsarAdmin admin = createAdminClient()) {
	doCeateOrModifyTopicsIfNeeded(topics);
} catch (Exception e) {
	throw new IllegalStateException("Could not create PulsarAdmin", e);
}
```

Whatever the approach, we should go ahead and make sure this is closed eventhough it is an exceptional condition on startup only.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationBadContextTests.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
+import java.util.Map;
+import org.apache.pulsar.client.admin.PulsarAdmin;
+import org.junit.jupiter.api.Test;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+
+public class PulsarAdministrationBadContextTests extends AbstractContainerBaseTests {
```
onobc, line 12: I think we can delete this now since you moved this into the other test

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTopicTests.java
onobc: There is a bit of validation in `getComponents`.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTopicTests.java
```diff
@@ -0,0 +1,52 @@
...
+		PulsarTopicBuilder builder = PulsarTopic.builder(topicName);
+		PulsarTopic topic = builder.build();
+
+		assertThat(topic.topicName()).isEqualTo(topicName);
+		assertThat(topic.numberOfPartitions()).isEqualTo(0);
+	}
+
+	@ParameterizedTest
+	@MethodSource("testTopicComponentsProvider")
+	void testTopicComponents(PulsarTopic topic, TopicDomain domain, String tenant, String namespace, String topicName) {
```
onobc, line 25: **[TinyNit]** One of the cool things w/ Junit5 was no longer needing to name test cases w/ the prefix `test`. It seems minute but it does clean things up a bit. Do you mind removing the renaming the tests? I can also do this in follow-on commit. Not trying to nit-pick you to death 😸

onobc (commented):
@alpreu I added a few minor comments to address but other than that, the code side of this is complete. 

I think the final things are: 
* Update docs (as you mentioned earlier)
* Consider adding this to one or both sample apps

Thanks

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationBadContextTests.java
alpreu: Oops, missed that

onobc:
> We should also make the `PulsarAdministration` bean auto-configured.

@sobychacko I would be in favor of doing an immediate follow-on PR for the admin auto-configuration. It will require property updates and tests etc..

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministrationOperations.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.core;
+
+/**
+ * The Pulsar administration contract.
+ *
+ * @author Chris Bono
```
onobc, line 22: Please keep yourself as the single author. We collaborated on it but this is your work my friend. :)

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/4bab93e744d13449bf83747caa7c205f986b2d59

PR #66, opened 2022-08-15
PulsarTemplate schema detection improvements

sobychacko:
PulsarTemplate can infer schema type for basic primitive types.
For complex types like JSON, AVRO etc, it needs to support
the ability for the user to provide the schema information.

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
@@ -46,6 +46,8 @@ public class PulsarTemplate<T> implements PulsarOperations<T> {
 
 	private final List<ProducerInterceptor> interceptors;
 
+	private Schema<T> schema;
+
 	/**
 	 * Constructs a template instance.
 	 * @param producerFactory the producer factory used to create the backing Pulsar
@@ -103,8 +105,16 @@ public class PulsarTemplate<T> implements PulsarOperations<T> {
 
 	private Producer<T> prepareProducerForSend(String topic, T message, MessageRouter messageRouter)
 			throws PulsarClientException {
-		Schema<T> schema = SchemaUtils.getSchema(message);
+		Schema<T> schema = this.schema != null ? this.schema : SchemaUtils.getSchema(message);
 		return this.producerFactory.createProducer(topic, schema, messageRouter, this.interceptors);
+	}
+
+	/**
+	 * Setter for schema.
+	 * @param schema provides the {@link Schema} used on this template
+	 */
+	public void setSchema(Schema<T> schema) {
+		this.schema = schema;
 	}
 
 }
--- a/spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
+++ b/spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
@@ -29,6 +29,7 @@ import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
@@ -45,6 +46,7 @@ import org.apache.pulsar.client.api.TopicMetadata;
 import org.apache.pulsar.client.api.interceptor.ProducerInterceptor;
 import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Named;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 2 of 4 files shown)


spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
```diff
@@ -118,6 +120,27 @@ void interceptorInvocationTest(String topic, List<ProducerInterceptor> intercept
...
+					.subscriptionName("test-specific-schema-subscription").subscribe()) {
+				PulsarProducerFactory<Foo> producerFactory = new DefaultPulsarProducerFactory<>(client,
+						Collections.singletonMap("topicName", topic));
+				PulsarTemplate<Foo> pulsarTemplate = new PulsarTemplate<>(producerFactory);
+				pulsarTemplate.setSchema(Schema.JSON(Foo.class));
+				Foo foo = new Foo("Foo-" + UUID.randomUUID(), "Bar-" + UUID.randomUUID());
+				pulsarTemplate.send(foo);
+
+				CompletableFuture<Message<Foo>> receiveMsgFuture = consumer.receiveAsync();
+				Message<Foo> msg = receiveMsgFuture.get(3, TimeUnit.SECONDS);
```
onobc, line 137: Have you seen AssertJ support for `CompletableFuture`? It is pretty handy. Here are some [examples](https://github.com/assertj/assertj-examples/blob/main/assertions-examples/src/test/java/org/assertj/examples/CompletableFutureAssertionsExamples.java).

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
```diff
@@ -216,4 +239,11 @@ V doSend(PulsarTemplate<String> template, String topic, String msg,
 
 	}
 
+	record Foo(String foo, String bar) {
+		@Override
+		public String toString() {
+			return "Foo{" + "foo='" + this.foo + '\'' + ", bar='" + this.bar + '\'' + '}';
```
onobc, line 245: I believe `record` also provides a reasonable `toString`.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -103,8 +105,16 @@ public CompletableFuture<MessageId> sendAsync(String topic, T message,
 
 	private Producer<T> prepareProducerForSend(String topic, T message, MessageRouter messageRouter)
 			throws PulsarClientException {
-		Schema<T> schema = SchemaUtils.getSchema(message);
+		Schema<T> schema = this.schema != null ? this.schema : SchemaUtils.getSchema(message);
 		return this.producerFactory.createProducer(topic, schema, messageRouter, this.interceptors);
 	}
 
+	/**
```
onobc, line 112: I am a bit torn on having this stored at the template level. Up until now we have made all inputs on the send methods. However, I also understand that adding params to that API increases the permutations of variants of the API methods along w/ maintenance/testing of them. This is probably a field that will not be changed on ever send call (but rather infrequently) so it is probably fine here at the top level. 

In general, the things I don't like about this pattern are:
1. keeping state makes the PT not thread-safe. Multiple threads could trample each other schema setting
2. as an API user it is not clear to me by looking at the `send` APIs that I should/may need to also call `setSchema` prior to invoking. Then once I find that info out I have to wonder what is it currently set to? Then should I set it back to what it was before I set it (aka cleanup after myself).

sobychacko:
@onobc Pulsar is pretty strict when it comes to schema usage. Once you decide on a particular schema for a topic, then it is not allowed to publish to the same topic with a different schema. So, different `send` invocations cannot provide different schema types on a whim. You need to delete/remove the topic, then delete the schema, etc. to use a new schema. I think it is pretty safe to set the schema on `PulsarTemplate`.  I am unsure if there is a threading issue since the schema cannot be changed after setting it (well, you can, but then it will throw an error).

sobychacko:
For the most part, we will do the inference on the framework side, so the only time, they have to set this is when they are using `JSON`, `AVRO` etc. In this case, we can document it (its already part of the PR) so that the schema setting is not associated with the `send` API methods.

sobychacko:
@onobc I see the multi-threading issue now since the template can be used across multiple topics, thus changing the schema. Good point.

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
```diff
@@ -134,9 +134,7 @@ void sendMessageWithSpecificSchemaTest() throws Exception {
 				pulsarTemplate.send(foo);
 
 				CompletableFuture<Message<Foo>> receiveMsgFuture = consumer.receiveAsync();
-				Message<Foo> msg = receiveMsgFuture.get(3, TimeUnit.SECONDS);
-
-				assertThat(msg.getValue()).isEqualTo(foo);
+				assertThat(receiveMsgFuture).isCompleted().isCompletedWithValueMatching(m -> m.getValue().equals(foo));
```
onobc, line 137: Nice. A total **nit** is that I think `isCompletedWithValueMatching` implies the `isCompleted` and we only need the former.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: We discussed offline and will explore a fluent send API in the near future. For now this approach is simple and works.

onobc (approved):

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
onobc: I will adjust during merge.

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/624dde15fcab669a1dc5175cee95f559bff7705f

PR #71, opened 2022-08-26
Add .sdkmanrc with 17.0.4-tem

eddumelendez:
`.sdkmanrc` is used by sdkman to automatically switch to the version
declared in the file.


onobc:
Thanks for adding this @eddumelendez . I went ahead and merged. The `CI PR` workflow was 🟥 but it is unrelated and I think @sobychacko is knocking some kinks out of the concurrency consumer tests.

Issue #50, opened 2022-08-02
Document config properties


onobc:
To take full advantage of the recently added [automatically generated config props docs](https://github.com/spring-projects-experimental/spring-pulsar/commit/b5d6365bdf78ebc2c55888f6f2bbc452a1759c6c), we should ensure that our configuration properties are documented as described [here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.developing-auto-configuration.custom-starter.configuration-keys).

onobc:
Take this time to also sneak in a minor update to the `PulsarOperation` blocking APIs that talk about how long they block etc..

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/67cb57c800a8cca0f6bc460ddd66af34c02ebcfa 

Thanks @alpreu !

PR #72, opened 2022-08-26
Followup "GH-41 Add auto-created topics"

alpreu:
This PR adds auto-configuration and documentation for #64.

### DIFF
--- a/spring-pulsar-docs/src/main/asciidoc/application-properties.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/application-properties.adoc
@@ -15,3 +15,5 @@ include::application-properties/pulsar-client.adoc[]
 include::application-properties/pulsar-producer.adoc[]
 
 include::application-properties/pulsar-consumer.adoc[]
+
+include::application-properties/pulsar-administration.adoc[]
--- a/spring-pulsar-sample-apps/src/main/java/app1/SpringPulsarBootApp.java
+++ b/spring-pulsar-sample-apps/src/main/java/app1/SpringPulsarBootApp.java
@@ -29,8 +29,10 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
 import org.springframework.pulsar.annotation.PulsarListener;
+import org.springframework.pulsar.core.PulsarAdministration;
 import org.springframework.pulsar.core.PulsarProducerFactory;
 import org.springframework.pulsar.core.PulsarTemplate;
+import org.springframework.pulsar.core.PulsarTopic;
 
 @SpringBootApplication
 public class SpringPulsarBootApp {
@@ -108,12 +110,34 @@ public class SpringPulsarBootApp {
 	}
 
 	/*
+	 * Create a partitioned topic using PulsarAdministration and then publish to the topic
+	 * and consume from it.
+	 */
+	@Bean
+	ApplicationRunner runner4(PulsarTemplate<String> pulsarTemplate, PulsarAdministration pulsarAdministration) {
+		String topicName = "hello-pulsar-partitioned-4";
+		PulsarTopic partitionedTopic = PulsarTopic.builder(topicName).numberOfPartitions(3).build();
+		pulsarAdministration.createOrModifyTopics(partitionedTopic);
+
+		return args -> {
+			for (int i = 0; i < 10; i++) {
+				pulsarTemplate.send(topicName, "This is message " + (i + 1));
+			}
+		};
+	}
+
+	@PulsarListener(subscriptionName = "subscription-4", topics = "hello-pulsar-partitioned-4")
+	void listen4(String message) {
+		this.logger.info("Message received from partitioned-topic : " + message);
+	}
+
+	/*
 	 * Publish and then use PulsarListener in batch listening mode.
 	 */
 	@Bean
-	ApplicationRunner runner4(PulsarProducerFactory<Foo> producerFactory) {
+	ApplicationRunner runner5(PulsarProducerFactory<Foo> producerFactory) {
 
-		String topic = "hello-pulsar-exclusive-4";
+		String topic = "hello-pulsar-exclusive-5";
 		PulsarTemplate<Foo> pulsarTemplate = new PulsarTemplate<>(producerFactory);
 		pulsarTemplate.setSchema(Schema.JSON(Foo.class));
 		return args -> {
@@ -124,9 +148,9 @@ public class SpringPulsarBootApp {
 		};
 	}
 
-	@PulsarListener(subscriptionName = "subscription-4", topics = "hello-pulsar-exclusive-4",
+	@PulsarListener(subscriptionName = "subscription-5", topics = "hello-pulsar-exclusive-5",
 			schemaType = SchemaType.JSON, batch = true)
-	void listen4(List<Foo> messages) {
+	void listen5(List<Foo> messages) {
 		this.logger.info("records received :" + messages.size());
 		for (Foo message : messages) {
 			this.logger.info("record : " + message);
--- a/spring-pulsar-sample-apps/src/main/java/app1/SpringPulsarBootApp.java
+++ b/spring-pulsar-sample-apps/src/main/java/app1/SpringPulsarBootApp.java
@@ -29,7 +29,6 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
 import org.springframework.pulsar.annotation.PulsarListener;
-import org.springframework.pulsar.core.PulsarAdministration;
 import org.springframework.pulsar.core.PulsarProducerFactory;
 import org.springframework.pulsar.core.PulsarTemplate;
 import org.springframework.pulsar.core.PulsarTopic;
@@ -114,14 +113,15 @@ public class SpringPulsarBootApp {
 	 * and consume from it.
 	 */
 	@Bean
-	ApplicationRunner runner4(PulsarTemplate<String> pulsarTemplate, PulsarAdministration pulsarAdministration) {
-		String topicName = "hello-pulsar-partitioned-4";
-		PulsarTopic partitionedTopic = PulsarTopic.builder(topicName).numberOfPartitions(3).build();
-		pulsarAdministration.createOrModifyTopics(partitionedTopic);
+	PulsarTopic partitionedTopic4() {
+		return PulsarTopic.builder("hello-pulsar-partitioned-4").numberOfPartitions(3).build();
+	}
 
+	@Bean
+	ApplicationRunner runner4(PulsarTemplate<String> pulsarTemplate) {
 		return args -> {
 			for (int i = 0; i < 10; i++) {
-				pulsarTemplate.send(topicName, "This is message " + (i + 1));
+				pulsarTemplate.send("hello-pulsar-partitioned-4", "This is message " + (i + 1));
 			}
 		};
 	}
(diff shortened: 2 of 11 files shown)


spring-pulsar-docs/src/main/asciidoc/application-properties.adoc
```diff
@@ -15,3 +15,5 @@ include::application-properties/pulsar-client.adoc[]
 include::application-properties/pulsar-producer.adoc[]
 
 include::application-properties/pulsar-consumer.adoc[]
+
+include::application-properties/pulsar-administration.adoc[]
```
alpreu, line 19: Here I'm not sure how these docs are generated, I just added the stub for consistency

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -59,6 +60,8 @@ public class PulsarProperties {
 
 	private final Producer producer = new Producer();
 
+	private final Administration administration = new Administration();
```
onobc, line 63: I am in favor of `spring.pulsar.admin` vs. `spring.pulsar.administration`. The reason is that it is much easier to type when doing props and it is just as informative (ie we don't lose anything by chopping it down). We could achieve that by naming this var as such. Wdyt?

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -59,6 +60,8 @@ public class PulsarProperties {
...
+
 	public Map<String, Object> buildConsumerProperties() {
```
onobc, line 65: **[TinyNit]** I think these got a little out of order before this  PR, but do you mind putting all of the `get<THING>` before the `build<THING>`s?

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -835,6 +842,186 @@ public void setSchemaType(SchemaType schemaType) {
...
+			map.from(this::isTlsHostnameVerificationEnable).to(properties.in("tlsHostnameVerificationEnable"));
+			map.from(this::isUseKeyStoreTls).to(properties.in("useKeyStoreTls"));
+			map.from(this::getSslProvider).to(properties.in("sslProvider"));
+			map.from(this::getTlsTrustStoreType).to(properties.in("tlsTrustStoreType"));
+			map.from(this::getTlsTrustStorePath).to(properties.in("tlsTrustStorePath"));
+			map.from(this::getTlsTrustStorePassword).to(properties.in("tlsTrustStorePassword"));
+			map.from(this::getTlsCiphers).to(properties.in("tlsCiphers"));
+			map.from(this::getTlsProtocols).to(properties.in("tlsProtocols"));
+			map.from(this::getConnectionTimeoutMs).to(properties.in("connectionTimeoutMs"));
+			map.from(this::getRequestTimeoutMs).to(properties.in("requestTimeoutMs"));
```
onobc, line 1016: I notice that there are a handful of properties that are 1st class props on the `PulsarAdminImpl` that are **also** in the config data map. It seems that some of the ones in the map are not used and instead only the ones directly on the admin impl are used. 

```java
this.connectTimeout = connectTimeout;
this.connectTimeoutUnit = connectTimeoutUnit;
this.readTimeout = readTimeout;
this.readTimeoutUnit = readTimeoutUnit;
this.requestTimeout = requestTimeout;
this.requestTimeoutUnit = requestTimeoutUnit;
```

The 1st class ones are only set using the values from the AdminBuilder, not the map of props that we are using. The issue I see is that users may try to set the requestTimeout but then will be disappointed when it does not actually get used. 

On the regular `PulsarClient` side of things this is not an issue as the conf map seems to be the final arbiter of chosen values.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -835,6 +842,186 @@ public void setSchemaType(SchemaType schemaType) {
...
+		public void setRequestTimeoutMs(long requestTimeoutMs) {
+			this.requestTimeoutMs = requestTimeoutMs;
+		}
+
+		public Map<String, Object> buildProperties() {
+			PulsarProperties.Properties properties = new Properties();
+
+			PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
+
+			map.from(this::getServiceUrl).to(properties.in("serviceUrl"));
```
onobc, line 1002: So the admin client and regular client both take in the `org.apache.pulsar.client.impl.conf.ClientConfigurationData` and have almost identical props here in our configuration of them. 

**Question 1**: Should they support the same exact properties? 
If so, we could consolidate some of the code w/ a common client props base class.

**Question 2**: Should each one support all of the properties in the clientconfigdata? 
if not, what is the final arbiter on which client support which options? I would think that the pulsar AdminBuilder / ClientBuilder provides that contract w/ its builder methods. 

If the builders are the contract, then one thought would be to populate their builders rather than the raw map. 


Whatever we decide here, we should circle back through the regular client properties @sobychacko and make sure we have them all covered. I do see some missing ones at 1st glance: 
```java
numListenerThreads
maxLookupRedirects
authParamMap
socks5ProxyAddress/Username/Password
```

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
```diff
@@ -90,7 +91,7 @@ void defaultBeansAreAutoConfigured() {
 						.hasSingleBean(PulsarTemplate.class).hasSingleBean(PulsarConsumerFactory.class)
 						.hasSingleBean(ConcurrentPulsarListenerContainerFactory.class)
 						.hasSingleBean(PulsarListenerAnnotationBeanPostProcessor.class)
-						.hasSingleBean(PulsarListenerEndpointRegistry.class));
+						.hasSingleBean(PulsarListenerEndpointRegistry.class).hasSingleBean(PulsarAdministration.class));
 	}
 
 	@Test
```
onobc, line 97: We should probably add a test that a custom admin client is respected. Simiilar to `customPulsarClientFactoryBeanIsRespected`

spring-pulsar-sample-apps/src/main/java/app1/SpringPulsarBootApp.java
```diff
@@ -107,13 +109,35 @@ void listen3(Foo message) {
 		this.logger.info("Message received :" + message);
 	}
 
+	/*
+	 * Create a partitioned topic using PulsarAdministration and then publish to the topic
+	 * and consume from it.
+	 */
+	@Bean
+	ApplicationRunner runner4(PulsarTemplate<String> pulsarTemplate, PulsarAdministration pulsarAdministration) {
```
onobc, line 117: @alpreu I think we can replace all of this w/ a `@Bean PulsarTopic` now that the auto-config is in place.

onobc (changes requested):
@alpreu I have some minor comments and then the concern of which props we should support (admin and regular client). But other than that, looks great and is ready to go. ~Lets come to conclusion on the properties and then we can get this one merged.~ Docs look great. Thanks for adding. 

**EDITED** I put a ~strike~ through the above concern based on [this comment](https://github.com/spring-projects-experimental/spring-pulsar/pull/72#discussion_r956441504).

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
sobychacko: +1 for `spring.pulsar.admin`.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
onobc: Me and @sobychacko discussed offline and came to the conclusion to make the `ClientConfigurationData` the source of truth for the supported properties for both admin and regular pulsar client. 

Let's also go w/ this set of mapped properties for the initial offering. We will create a follow-on issue to revisit and fine-tune the properties to be sure we are exposing the ones we want for both admin and regular client.

@alpreu please resolve this comment once you have digested it.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: Good catch! This is an oversight on our part, I'm opening a bug ticket for the builder.
For now I will remove the timeouts from the PulsarProperties to prevent this issue.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: Following are some thoughts I have on this. The `ClientConfigurationData` being used for both the admin as well as the regular client is a misuse on our side. 
First, not all properties are shared between both clients, for example, the SOCKS5 proxy properties are only used on the regular client. 
Further, there are properties that even though they are shared, behave differently for the clients. An example of this is the `serviceUrl`, which for the regular client will look like `pulsar+ssl://localhost:6650`, while for the admin it will look like `https://localhost:8080`.
This makes re-using the same set of properties for both difficult.

onobc: By `source of truth for the supported properties for both admin and regular pulsar client` I really just mean these are the properties we would support. Thanks for the info on their divergence (admin and regular client) @alpreu .

In this case, I bet the actual AdminBuilder and PulsarClientBuilder are the actual final arbiter of what is supported. It feels to me like we should support whatever their respective builders support. 

For this PR I think what we have is fine for 1st pass. We should create a follow on ticket to explore this a bit more.

onobc:
@alpreu thanks for adding the all of the config props docs (this takes care of #50 as well). I will review the updates here shortly and get this merged in.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
onobc: Will address in #76

onobc (approved):
Thanks again for another awesome contribution @alpreu !

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: Yes, it is a good idea to go through the properties again. 
I also filed a bug ticket in Pulsar for the properties that get overridden by the builder's default values.

onobc: > Yes, it is a good idea to go through the properties again. I also filed a bug ticket in Pulsar for the properties that get overridden by the builder's default values.

Thank you for also filing the bug ticket in Pulsar.

Issue #57, opened 2022-08-09
Tests are failing when running on M1 Mac


onobc:
**Steps to reproduce:** Run `gradle clean build` on an M1 Mac.

**Possible cause:** Pulsar image running super slow because of M1 Mac?

**Workaround:** Modify the image used in https://github.com/spring-projects-experimental/spring-pulsar/blob/749f2dbb5e386f77e842af2ece547fbdc0e8f84f/spring-pulsar/src/test/java/org/springframework/pulsar/core/AbstractContainerBaseTests.java#L24
with 
```java
static final DockerImageName PULSAR_IMAGE = DockerImageName.parse("kezhenxu94/pulsar").asCompatibleSubstituteFor("apachepulsar/pulsar");
```
### Work items

1. Understand what the root cause is. What is in the alternative image that makes it work? Is there some other change we should be considering or is this just the "M1 support in general is still catching up" category?
2. Temporary stop-gap so that other users do not run into this, we could implement the above workaround automatically via Junit5 [@EnabledOnOs](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-os) annotation (one for M1 and one for all others).

alpreu:
I believe the root cause is that the standard Pulsar image running through emulation is very slow. This leads to Testcontainers startup check to time out, thus stopping further test execution.

The issue around offering ARM based images can be found here: 
https://github.com/apache/pulsar/issues/12944

And the experimental image above was built using this PR: 
https://github.com/apache/pulsar/pull/14005

I tried to re-ignite the discussion on the mailing list around this topic and looped in colleagues as well.

For the time being, I think it's a good idea to enable the workaround using the JUnit annotation as you described above.

onobc:
**NOTE:** As of 2.11.0 the standard Pulsar image is now able to run on Mac M1.

PR #75, opened 2022-08-29
Detect M1 mac in test and substitute container image (#57)

alpreu:
This is a best-effort solution for issue #57.

It uses the JVMs system properties to detect the M1 architecture.
This solution will fail if the JVM is running through Rosetta emulation. Other approaches that are independent of the JVM suffer similar problems (see [this question about detecting the M1 from the cli](https://stackoverflow.com/questions/65259300/detect-apple-silicon-from-command-line) for example) so I believe there is no really good solution for this edge case.

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AbstractContainerBaseTests.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AbstractContainerBaseTests.java
@@ -16,16 +16,17 @@
 
 package org.springframework.pulsar.autoconfigure;
 
+import java.util.Locale;
+
 import org.testcontainers.containers.PulsarContainer;
 import org.testcontainers.utility.DockerImageName;
 
 abstract class AbstractContainerBaseTests {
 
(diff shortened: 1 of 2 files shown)


spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AbstractContainerBaseTests.java
```diff
@@ -38,4 +39,18 @@ protected static String getHttpServiceUrl() {
 		return PULSAR_CONTAINER.getHttpServiceUrl();
 	}
 
+	private static boolean isRunningOnMacM1() {
```
onobc, line 42: > It uses the JVMs system properties to detect the M1 architecture.
This solution will fail if the JVM is running through Rosetta emulation. Other approaches that are independent of the JVM suffer similar problems (see [this question about detecting the M1 from the cli](https://stackoverflow.com/questions/65259300/detect-apple-silicon-from-command-line) for example) so I believe there is no really good solution for this edge case.

I think this is a pragmatic choice. We could later provide an env var that users in the "rosetta" group could set as an override/last resort. Again, I think this is fine for now. 


If we start needing to do this same logic in other places we could add something like `MacM1Detector` that encapsulates how we determine this info.

onobc (commented):
Looks good @alpreu !

PR #82, opened 2022-09-02
Add AVRO and (inline) KeyValue support for PulsarListener

alpreu:
This PR adds support for Avro and KeyValue types in PulsarListener.

Currently, this only supports inline KeyValue types, as we cannot detect which KeyValueEncodingType is used.

Still missing documentation.

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
@@ -20,13 +20,16 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
 
 import org.apache.commons.logging.LogFactory;
 import org.apache.pulsar.client.api.Consumer;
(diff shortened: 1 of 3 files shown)


onobc:
@alpreu, 

Thanks for another awesome PR!. Me and @sobychacko reviewed it and it looks good. I know you still wanted to add docs but Soby was going to be working in the same area over the weekend so to avoid the conflict/merge we went ahead and merged it in. 

You will notice I added a [polish commit](
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/597bacde4125585c7bfbac1ded50ba73a25f745e). This was aimed at the duplication that already existed in `SchemaUtils` as well as in the `MethodPulsarListenerEndpoint`. The only part of the newly added code I changed was to rename from "method"Parameter to "message"Parameter which helped me (coming in fresh to this particular section of the code) to more easily resonate w/ what the code was doing - it made it less generic/abstract to follow along with. 

I think we may want to also add a couple of tests to cover `List<User>` and `List<>` (not sure if the latter is a valid choice that we allow though. 

Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/c2732c6ee602792605d4aafafc1ddef708ea1946

PR #87, opened 2022-09-05
Followup complex schemas

alpreu:
This is the followup PR for #82, adding tests for containing parameters and updating the docs with the supported schema types.


alpreu:
Regarding the last commit (protobuf), I liked the idea of using the gradle plugin to generate the example java class from a .proto file during the build time, but after fighting with it for quite some time I gave up on it.

onobc:
Thanks for the follow up @alpreu . I will review this sometime today. I will also see if I can get the protoc going from the plugin (I have had success w/ this recently on an unrelated project).

spring-pulsar/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
 	api ('org.springframework.retry:spring-retry') {
 		exclude group: 'org.springframework'
 	}
+	api 'com.google.protobuf:protobuf-java:3.21.5'
```
onobc, line 18: A couple of things on this new addition...

1. I was about to comment to suggest making this dep. optional as I thought that it would bring in `Guava`. **However**, I am glad that is not the case as it is only a `test` [dependency](https://mvnrepository.com/artifact/com.google.protobuf/protobuf-java/3.21.5). I decided to keep this comment in order to say out loud that if we ever **have** to bring Guava in that we should shade it. It has caused me grief as a transitive dependency in the past. 




2. We have the versions managed via a Gradle Java platform in `spring-pulsar-dependencies`. Do you mind removing the version number here and then add a constraint for it in platform? This will look just like the one for [caffeine](https://github.com/spring-projects-experimental/spring-pulsar/blob/cc3842062e970ae4b64ade537d085bd0ec309ac9/spring-pulsar-dependencies/build.gradle#L39). I have been meaning to add an "Adding new dependencies" section for this in the [Working with the code Wiki](https://github.com/spring-projects-experimental/spring-pulsar/wiki/Working-with-the-Code) but have not gotten around to it yet.

onobc (changes requested):
@alpreu another excellent contribution (what else would I expect at this point 😸 ). Thanks for beefing up the tests and docs and also the protobuf schema support. 

My only comment is around dep. version management. Once that is addressed we can get this merged. I will check when I come online tomorrow. 

As for the proto test file generation via the proto plugin... I did not get a chance to play w/ it. But we should be able to figure it out - not super important though. IIRC the plugin is picky on where the proto lives, and expects it to be in `src/[main|test]/protobuf` (not 100% sure though).

alpreu:
Thanks for the kind words Chris and thanks for drawing my attention to how you manage the dependencies.
I moved the version there and also saw that you tried to keep everything neatly ordered, so I reordered the outliers as well while I was at it ;)

onobc (approved):
Thanks for the update @alpreu . 

> and also saw that you tried to keep everything neatly ordered, so I reordered the outliers as well while I was at it ;)

Much appreciated!. Merging now.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -275,22 +285,42 @@ void avroSchema() throws Exception {
...
 		void keyvalueSchema() throws Exception {
 			PulsarProducerFactory<KeyValue<String, Integer>> pulsarProducerFactory = new DefaultPulsarProducerFactory<>(
 					pulsarClient, Collections.emptyMap());
 			PulsarTemplate<KeyValue<String, Integer>> template = new PulsarTemplate<>(pulsarProducerFactory);
-
 			Schema<KeyValue<String, Integer>> kvSchema = Schema.KeyValue(Schema.STRING, Schema.INT32,
 					KeyValueEncodingType.INLINE);
-
 			template.setSchema(kvSchema);
-			template.send("keyvalue-topic", new KeyValue<>("Kevin", 3));
```
onobc, line 292: @alpreu I forgot to mention this... I see what you did here w/ the clever naming of the message payload to alpha-relate to the feature being tested (avro/avi, json/jason, keyvalue/kevin, protobuf/paul). Seems minor thing but its much better than "foo" or "msg-N" as it relates to test and could help in associating during debugging later etc..

Issue #79, opened 2022-09-01
Express deadLetterPolicy as a bean on PulsarListener


sobychacko:
`deadLetterPolicy` on Pulsar Consumer is an implementation of `DeadLetterPolicy`. Applications should be able to provide a bean for `DeadLetterPolicy` and express that on `PulsarListener` as a bean name. The pattern described [here](https://github.com/spring-projects-experimental/spring-pulsar/issues/78#issue-1358134156) can be used as a blueprint.

sobychacko:
Note that `deadLetterPolicy` can only be specified for shared subscription types. On exclusive and failover subscriptions, it does not have any effect.

PR #88, opened 2022-09-06
Add deadLetterPolicy to PulsarListener

alpreu:
This PR adds the ability to add DeadLetterPolicy beans to the PulsarListener annotation (#79).
Mostly based on Soby's work done in #85

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListener.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListener.java
@@ -167,15 +167,15 @@ public @interface PulsarListener {
 	 * The bean name or a 'SpEL' expression that resolves to a
 	 * {@link org.apache.pulsar.client.api.RedeliveryBackoff} to use on the consumer to
 	 * control the redelivery backoff of messages after a negative ack.
-	 * @return the bean name or empty string to not set the backoff
+	 * @return the bean name or empty string to not set the backoff.
 	 */
 	String negativeAckRedeliveryBackoff() default "";
 
 	/**
 	 * The bean name or a 'SpEL' expression that resolves to a
 	 * {@link org.apache.pulsar.client.api.DeadLetterPolicy} to use on the consumer to
-	 * configure a maximum message redelivery.
-	 * @return the bean name or empty string to not set any dead letter policy
+	 * configure a dead letter policy for message redelivery.
+	 * @return the bean name or empty string to not set any dead letter policy.
 	 */
 	String deadLetterPolicy() default "";
 
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListener.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListener.java
@@ -46,6 +46,7 @@ import org.springframework.pulsar.config.PulsarListenerEndpointRegistry;
  *
  * @author Soby Chacko
  * @author Chris Bono
+ * @author Alexander Preuß
  */
 @Target({ ElementType.TYPE, ElementType.METHOD, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
@@ -169,5 +170,13 @@ public @interface PulsarListener {
 	 * @return the bean name or empty string to not set the backoff
 	 */
 	String negativeAckRedeliveryBackoff() default "";
+
+	/**
+	 * The bean name or a 'SpEL' expression that resolves to a
+	 * {@link org.apache.pulsar.client.api.DeadLetterPolicy} to use on the consumer to
+	 * configure a maximum message redelivery.
+	 * @return the bean name or empty string to not set any dead letter policy
+	 */
+	String deadLetterPolicy() default "";
 
 }
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
@@ -79,9 +79,11 @@ public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T>
 		final Map<String, Object> properties = new HashMap<>(this.consumerConfig);
 		properties.putAll(propertiesToOverride);
 
+		// Remove deadLetterPolicy from the properties here and save it to re-apply after
+		// calling `loadConf` (https://github.com/apache/pulsar/issues/11646)
 		DeadLetterPolicy deadLetterPolicy = null;
 		if (properties.containsKey("deadLetterPolicy")) {
-			deadLetterPolicy = (DeadLetterPolicy) properties.remove("deadLetterPolicy"); // https://github.com/apache/pulsar/issues/11646
+			deadLetterPolicy = (DeadLetterPolicy) properties.remove("deadLetterPolicy");
 		}
 
 		if (!CollectionUtils.isEmpty(properties)) {
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import org.apache.pulsar.client.api.BatchReceivePolicy;
 import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.ConsumerBuilder;
+import org.apache.pulsar.client.api.DeadLetterPolicy;
 import org.apache.pulsar.client.api.PulsarClient;
 import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.client.api.RedeliveryBackoff;
@@ -36,6 +37,7 @@ import org.springframework.util.CollectionUtils;
  *
  * @param <T> underlying payload type for the consumer.
  * @author Soby Chacko
+ * @author Alexander Preuß
  */
 public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T> {
 
@@ -77,8 +79,17 @@ public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T>
 		final Map<String, Object> properties = new HashMap<>(this.consumerConfig);
 		properties.putAll(propertiesToOverride);
 
+		DeadLetterPolicy deadLetterPolicy = null;
+		if (properties.containsKey("deadLetterPolicy")) {
+			deadLetterPolicy = (DeadLetterPolicy) properties.remove("deadLetterPolicy"); // https://github.com/apache/pulsar/issues/11646
+		}
+
 		if (!CollectionUtils.isEmpty(properties)) {
 			consumerBuilder.loadConf(properties);
+		}
+
+		if (deadLetterPolicy != null) {
+			consumerBuilder.deadLetterPolicy(deadLetterPolicy);
 		}
 
 		if (properties.containsKey("negativeAckRedeliveryBackoff")) {
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
@@ -32,6 +32,7 @@ import java.util.stream.StreamSupport;
 
 import org.apache.pulsar.client.api.BatchReceivePolicy;
 import org.apache.pulsar.client.api.Consumer;
+import org.apache.pulsar.client.api.DeadLetterPolicy;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.MessageListener;
@@ -238,6 +239,10 @@ public class DefaultPulsarMessageListenerContainer<T> extends AbstractPulsarMess
 			final RedeliveryBackoff negativeAckRedeliveryBackoff = DefaultPulsarMessageListenerContainer.this.negativeAckRedeliveryBackoff;
 			if (negativeAckRedeliveryBackoff != null) {
 				currentProperties.put("negativeAckRedeliveryBackoff", negativeAckRedeliveryBackoff);
+			}
+			final DeadLetterPolicy deadLetterPolicy = DefaultPulsarMessageListenerContainer.this.deadLetterPolicy;
+			if (deadLetterPolicy != null) {
+				currentProperties.put("deadLetterPolicy", deadLetterPolicy);
 			}
 		}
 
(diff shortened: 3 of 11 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -77,10 +79,19 @@ public Consumer<T> createConsumer(Schema<T> schema, BatchReceivePolicy batchRece
...
+		}
+
 		if (!CollectionUtils.isEmpty(properties)) {
 			consumerBuilder.loadConf(properties);
 		}
 
+		if (deadLetterPolicy != null) {
+			consumerBuilder.deadLetterPolicy(deadLetterPolicy);
+		}
+
```
alpreu, line 94: Here we have to use this ugly workaround because `loadConf` cannot read in the DeadLetterPolicy correctly

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListener.java
```diff
@@ -170,4 +171,12 @@
 	 */
 	String negativeAckRedeliveryBackoff() default "";
 
+	/**
+	 * The bean name or a 'SpEL' expression that resolves to a
+	 * {@link org.apache.pulsar.client.api.DeadLetterPolicy} to use on the consumer to
+	 * configure a maximum message redelivery.
```
onobc, line 177: Close.... ;)

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -77,10 +79,19 @@ public Consumer<T> createConsumer(Schema<T> schema, BatchReceivePolicy batchRece
 		final Map<String, Object> properties = new HashMap<>(this.consumerConfig);
 		properties.putAll(propertiesToOverride);
 
+		DeadLetterPolicy deadLetterPolicy = null;
```
onobc, line 82: Thanks for adding link to underlying cause of this workaround. I think adding a one-liner (similar to your [review comment](https://github.com/spring-projects-experimental/spring-pulsar/pull/88#discussion_r963599760) may be helpful as it may not be clear (in say, 6mos. from now) that the code on 91 (after loadConf) is also part of the workaround.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -239,6 +240,10 @@ private void populateAllNecessaryPropertiesIfNeedBe(Map<String, Object> currentP
 			if (negativeAckRedeliveryBackoff != null) {
 				currentProperties.put("negativeAckRedeliveryBackoff", negativeAckRedeliveryBackoff);
```
onobc, line 241: **[UNRELATED]** Not directly related to this PR as it came to mind during the previous similar one.... these constants are starting to proliferate (across classes as well). Not sure where a good home would be for them as I do like having the constants "close to the code" but just something for us to all noodle on as we go back in here next time etc..

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
```diff
@@ -41,28 +42,30 @@
 
 /**
  * @author Soby Chacko
+ * @author Alexander Preuß
  */
 public class ConcurrentPulsarMessageListenerContainerTests {
 
 	@Test
 	@SuppressWarnings("unchecked")
```
onobc, line 50: **[NIT]** If we end up having the same suppression on all of the test methods, may just bring it up to the class level. Wdyt?

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
```diff
@@ -105,6 +96,17 @@ void basicConcurrencyTesting() throws Exception {
...
 	void exclusiveSubscriptionMustUseSingleThread() throws Exception {
+		MockEnvironment env = setupMockEnvironment(SubscriptionType.Exclusive);
+		ConcurrentPulsarMessageListenerContainer<String> concurrentContainer = env.concurrentContainer();
+
+		concurrentContainer.setConcurrency(3);
+
+		assertThatThrownBy(concurrentContainer::start).isInstanceOf(IllegalStateException.class)
```
onobc, line 104: Nice use/test.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
```diff
@@ -105,6 +96,17 @@ void basicConcurrencyTesting() throws Exception {
 	@Test
 	@SuppressWarnings("unchecked")
 	void exclusiveSubscriptionMustUseSingleThread() throws Exception {
+		MockEnvironment env = setupMockEnvironment(SubscriptionType.Exclusive);
```
onobc, line 99: **[NIT]** I do like the addition of this aggregate mock holder. I would consider renaming it to something other than "Environment" as that has a strong mental tie to the Spring Environment. You know, naming... the hardest thing to do.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
```diff
@@ -202,4 +204,60 @@ void negativeAckRedeliveryBackoff() throws Exception {
...
+		dlqContainerProperties.setSubscriptionType(SubscriptionType.Shared);
+		dlqContainerProperties.setTopics(new String[] { "dlq-topic" });
+		DefaultPulsarMessageListenerContainer<String> dlqContainer = new DefaultPulsarMessageListenerContainer<>(
+				pulsarConsumerFactory, dlqContainerProperties);
+		dlqContainer.start();
+
+		PulsarContainerProperties pulsarContainerProperties = new PulsarContainerProperties();
+		pulsarContainerProperties.setMessageListener((PulsarRecordMessageListener<?>) (consumer, msg) -> {
+			latch.countDown();
+			if (((String) msg.getValue()).endsWith("4")) {
```
onobc, line 237: Another alternative would be to check the latch current count instead of the message payload.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
```diff
@@ -202,4 +204,60 @@ void negativeAckRedeliveryBackoff() throws Exception {
...
+		final PulsarTemplate<String> pulsarTemplate = new PulsarTemplate<>(pulsarProducerFactory);
+		for (int i = 0; i < 5; i++) {
+			pulsarTemplate.send("hello john doe" + i);
+		}
+
+		// DLQ consumer should receive 1 msg
+		assertThat(dlqLatch.await(10, TimeUnit.SECONDS)).isTrue();
+		// Normal consumer should receive 5 msg + 1 re-delivery
+		assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
+		container.stop();
```
onobc, line 259: These may not always be closed. The other tests in here have the same issue though. I am fine if we clean that up in a follow on PR or commit.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -263,6 +264,46 @@ public RedeliveryBackoff redeliveryBackoff() {
 
 	}
 
+	@Nested
+	@ContextConfiguration(classes = DeadLetterPolicyTest.DeadLetterPolicyConfig.class)
+	class DeadLetterPolicyTest {
+
+		static CountDownLatch latch = new CountDownLatch(2);
```
onobc, line 271: **[NIT]** These may be able to be `private`.

onobc (changes requested):
@alpreu thanks for another awesome contribution! Looks great. I have a few minor comments.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Created an issue to track this: https://github.com/spring-projects-experimental/spring-pulsar/issues/90

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
alpreu: I just saw that due to extracting the setup method, only `basicConcurrencyTesting` still requires suppression. So either we have it on both that and the `setupMockEnvironment` method or on the class level. I'm fine with both

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
alpreu: The kudos belong to @sobychacko, it's just the git diff that makes it look like new code 😉

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
alpreu: I refactored it now to use an Integer message type and moved to one-based indexing, which improves readability quite a bit

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
alpreu: I was just thinking about another name and was going for MockComponents but I guess that doesn't really work either 😂

alpreu: Maybe PulsarMock? Sounds very generic but then again it's really just for holding the individual objects

onobc: Hehe. Maybe `PulsarListenerMockComponents`?

spring-pulsar/src/test/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainerTests.java
onobc: I prefer it fine-grain unless it is on ever single method. In this case I would prefer on those 2 places (setupMock + basicCon..). And, truly, a total NIT either way. 😄

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListener.java
onobc: Hehe. Sorry...
```
to use on the consumer to configure a maximum message redelivery..
```
As in "maximum message redelivery" is close to "dead letter policy"

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -268,8 +268,8 @@ public RedeliveryBackoff redeliveryBackoff() {
 	@ContextConfiguration(classes = DeadLetterPolicyTest.DeadLetterPolicyConfig.class)
 	class DeadLetterPolicyTest {
 
-		static CountDownLatch latch = new CountDownLatch(2);
-		static CountDownLatch dlqLatch = new CountDownLatch(1);
+		private static final CountDownLatch latch = new CountDownLatch(2);
```
onobc, line 271: I brought this up in a [PR recently](https://github.com/spring-projects-experimental/spring-pulsar/pull/85#discussion_r962250470) w/ @sobychacko and in other Spring projects we do not use `final` everywhere. I think we should stay consistent w/ that approach here as well. 

For now, let's not add `final` until we decide what we want to go w/.  Also, whatever we decide we should probably add/configure the format/checkstyle to enforce it.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
onobc: Thanks for adding the comment in the code as well @alpreu

onobc (approved):

onobc:
I have fixed conflict locally but when I run all tests I get a failure in PulsarListenerTests (of course works fine when run by itself). 

```java
PulsarListenerTests > DeadLetterPolicyTest STANDARD_OUT
    12:29:31.220 [Test worker] ERROR org.springframework.pulsar.listener.DefaultPulsarMessageListenerContainer - Pulsar client exceptions.
    org.apache.pulsar.client.api.PulsarClientException$IncompatibleSchemaException: {"errorMsg":"Incompatible schema: exists schema type INT32, new schema type STRING","reqId":2109641079268052675, "remote":"localhost/127.0.0.1:51249", "local":"/127.0.0.1:51436"}
        at org.apache.pulsar.client.api.PulsarClientException.unwrap(PulsarClientException.java:1014) ~[pulsar-client-api-2.10.1.jar:2.10.1]
        at org.apache.pulsar.client.impl.ConsumerBuilderImpl.subscribe(ConsumerBuilderImpl.java:101) ~[pulsar-client-2.10.1.jar:2.10.1]
        at org.springframework.pulsar.core.DefaultPulsarConsumerFactory.createConsumer(DefaultPulsarConsumerFactory.java:104) ~[main/:?]
        at org.springframework.pulsar.listener.DefaultPulsarMessageListenerContainer$Listener.<init>(DefaultPulsarMessageListenerContainer.java:191) ~[main/:?]
        at org.springframework.pulsar.listener.DefaultPulsarMessageListenerContainer.doStart(DefaultPulsarMessageListenerContainer.java:95) ~[main/:?]
        at org.springframework.pulsar.listener.AbstractPulsarMessageListenerContainer.start(AbstractPulsarMessageListenerContainer.java:169) ~[main/:?]
        at org.springframework.pulsar.listener.ConcurrentPulsarMessageListenerContainer.doStart(ConcurrentPulsarMessageListenerContainer.java:83) ~[main/:?]
        at org.springframework.pulsar.listener.AbstractPulsarMessageListenerContainer.start(AbstractPulsarMessageListenerContainer.java:169) ~[main/:?]
        at org.springframework.pulsar.config.PulsarListenerEndpointRegistry.startIfNecessary(PulsarListenerEndpointRegistry.java:239) ~[main/:?]
        at org.springframework.pulsar.config.PulsarListenerEndpointRegistry.start(PulsarListenerEndpointRegistry.java:191) ~[main/:?]
        at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:178) ~[spring-context-6.0.0-SNAPSHOT.jar:6.0.0-SNAPSHOT]
        at org.springframework.context.support.DefaultLifecycleProcessor$LifecycleGroup.start(DefaultLifecycleProcessor.java:356) ~[spring-context-6.0.0-SNAPSHOT.jar:6.0.0-SNAPSHOT]
        at java.lang.Iterable.forEach(Iterable.java:75) ~[?:?]
```

I am going to merge as @sobychacko is going to build on top of this work. @sobychacko can you follow up w/ the test failure?

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/25dcfab6070ded7d5a5386e8e661c669f91598f1

Issue #95, opened 2022-09-08
PHASE 1: Make client authentication simpler


onobc:
Typically, client authentication is configured by invoking the `authentication` method on the client builder and passing in the required info - for example:
```java
URL issuerUrl = new URL("https://auth.server.cloud/");
URL credentialsUrl = new URL("file:///path/to/KeyFile.json");
String audience = "urn:sn:acme:dev:my-instance";

PulsarClient client = PulsarClient.builder()
    .serviceUrl("pulsar://broker.example.com:6650/")
    .authentication(
        AuthenticationFactoryOAuth2.clientCredentials(issuerUrl, credentialsUrl, audience))
    .build();
```

However, the only current exposed mechanism to configure client authentication is via setting the `spring.pulsar.client.auth-*` properties. Moreover, to configure [oauth 2.0 access tokens](https://pulsar.apache.org/docs/2.10.1/security-oauth2/#java) based auth, the `spring.pulsar.client.auth-params` property must be set to a JSON string of all of the required values. It is cumbersome. 

```yaml
spring:
  pulsar:
    client:
      auth-plugin-class-name: org.apache.pulsar.client.impl.auth.oauth2.AuthenticationOAuth2
      #auth-params-map.issuer-url: https://auth.server.cloud/
      #auth-params-map.private-key: file:///Users/some-key.json
      #auth-params-map.audience: urn:sn:acme:dev:my-instance
      auth-params: "{\"privateKey\":\"file:///Users/some-key.json\",\"issuerUrl\":\"https://auth.server.cloud/", \"audience\":\"urn:sn:acme:dev:my-instance"}"
```

We want to improve DXP in this area by allowing users to configure the simpler `spring.pulsar.client.auth-params-map.*` properties OR if they so choose, the more complicated `spring.pulsar.client.auth-params` property. Rather than simply pass these properties down into the client builder we will instead use them to directly call the `authentication` method during auto-configuration. 

**NOTE:** 
We should probably remove the properties from the map before passing them down into the client builder to avoid the builder to attempt to do the same auth logic. However, it may check if the auth is not null before doing anything. Need to look into this,

alpreu:
I believe it's a good idea to remove `auth-params-map` completely as it's basically deprecated and some of the newer `Authentication` classes do not even implement this code path anymore. 
Auto-creating the `Authentication` object and passing it into the builder sounds like a good idea. 
[...]

PR #97, opened 2022-09-09
Convert auth-param-map parameters into auth-params enconded string

alpreu:
This PR implements the conversion of auth-param-map parameters into the auth-params encoded string (#95)

TODO:
* conversion only works if keys are provided in camelCase as `spring.pulsar.client.auth-param-map.myKey` but fails for `spring.pulsar.client.auth-param-map.my-key`
* rename `auth-param-map` to `authentication` to not confuse users with deprecated Pulsar API


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -35,6 +35,9 @@
 import org.apache.pulsar.client.api.SubscriptionInitialPosition;
 import org.apache.pulsar.client.api.SubscriptionType;
 import org.apache.pulsar.common.schema.SchemaType;
+import org.apache.pulsar.common.util.ObjectMapperFactory;
```
onobc, line 38: I like using Pulsar's OMF here as it will ideally be configured in a way that will produce the JSON that Pulsar expects to consume.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -35,6 +35,9 @@
...
+import org.apache.pulsar.shade.com.fasterxml.jackson.core.JsonProcessingException;
```
onobc, line 39: Lets use the non-shaded versions of these.

alpreu: That works for the `JsonProcessingException`, but the compiler complains when trying to replace the  shaded `ObjectMapper` coming from Pulsar's factory with the Jackson one

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1261,7 +1277,7 @@ public Map<String, Object> buildProperties() {
 			map.from(this::getListenerName).to(properties.in("listenerName"));
 			map.from(this::getAuthPluginClassName).to(properties.in("authPluginClassName"));
 			map.from(this::getAuthParams).to(properties.in("authParams"));
-			map.from(this::getAuthParamsMap).to(properties.in("authParamMap"));
+			map.from(convertToEncodedParamString(this.getAuthParamsMap())).to(properties.in("authParams"));
```
onobc, line 1280: I was also thinking in the case where the user has set `authParams` directly that they are wanting to be in control and in that case backoff the special processing.  I think as you have it currently they will be overwritten.  Wdyt?


_NOTE: Only pasting this because I had it drafted locally.. not saying "use my code instead"._ 

```java
// If authParams is set, leave as-is, user is in control.
if (StringUtils.hasText(this.getAuthParams())) {
  map.from(this::getAuthParams).to(properties.in("authParams"));
  map.from(this::getAuthParamsMap).to(properties.in("authParamMap"));
}
else if (!CollectionUtils.isEmpty(this.getAuthParamsMap())) { // Otherwise, build authParams from authParamsMap
  try {
    String authParamsJson = ObjectMapperFactory.create().writeValueAsString(this.authParamsMap);
    map.from(authParamsJson).to(properties.in("authParams"));
  }
  catch (JsonProcessingException e) {
    throw new IllegalStateException("Failed to convert authParamsMap to JSON string", e);
  }
}

```

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1438,7 +1454,7 @@ public Map<String, String> getAuthParamMap() {
 			return this.authParamMap;
 		}
 
-		public void setAuthParamMap(Map<String, String> authParamMap) {
+		public void setAuthParamMap(SortedMap<String, String> authParamMap) {
```
onobc, line 1457: If we implement the change I suggested [above](https://github.com/spring-projects-experimental/spring-pulsar/pull/97/files#r967247500) where user can set the `authParams` directly, does SB respect the sorted contract? I would expect so, but just not 100% sure.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
onobc: Ah ok. We can get rid of that by not holding a reference and just do `ObjectMapperFactory.create().writeValueAsString` inline.

onobc (changes requested):
@alpreu awesome work (as usual). 😄 Looks good. Couple of small comments to address. 

The final part is updating the ref docs w/ a section on auth and show what to set for each of the 6 auth types. If you have time to get to that cool - if not cool, I will gladly add that part in a follow on.  The description on the Github issue has a draft that can be used for this.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: In that case, the Exception type in the catch needs to be the shaded one again 😁

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: AFAIK the client code configures the Authentication with either `authParams` or `authParamsMap`, not both (I have to confirm this though). If that is the case, I think the current way of using the parameter map in case it is not empty seems fine to me.

onobc: If that is the case then, agreed. I think we should rename from `convert...` to `maybeConvert...` to indicate that it will do nothing in the empty case a bit more clearly. Nit though.

onobc:
@alpreu on the concern of

> conversion only works if keys are provided in camelCase as spring.pulsar.client.auth-param-map.myKey but fails for spring.pulsar.client.auth-param-map.my-key

Yep, the keys are case sensitive going into Pulsar so I think this is a realistic requirement for users . As you pointed out offline though, when set as env vars we lose that casing fidelity and it will always come in as lower-cased.

For example, `SPRING_PULSAR_CLIENT_AUTHPARAMSMAP_ISSUER_ID=foo` will come in as `spring.pulsar.client.authParamsMap.issuerid` (but Pulsar expects `issuerId`).

I think we can to do some canonicalization by having a set of well-known keys (`issuerid -> issuerId`) and replace if necessary during set/processing. The set of properties is small so that makes it more palatable. 


Oh, also if you have not seen this.... Junit Pioneer make env var testing "not suck" via...

```java
@SetEnvironmentVariable(key = "SPRING_PULSAR_CLIENT_AUTHPARAMSMAP_ISSUERID", value = "foo")
@Test 
void blahBlahBlah() {
  ....
}
```

`build.gradle`
```grovvy
testImplementation 'org.junit-pioneer:junit-pioneer:1.7.1'
```

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: Just confirmed it, the client implementation will exclusively use the String authParams when set, or fall back to the map: https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PulsarClientImpl.java#L253-L257

onobc: Good to know @alpreu . 

However, now that we have renamed from `authParamsMap -> authentication` I think this situation can now occur. User sets some `authParams` and then also some `authentication` params,, we will overwrite the former. I wonder if we should throw an exception if the user sets `authParams | authParamsMap` stating they should use `authentication` instead.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/AuthParameterUtils.java
```diff
@@ -0,0 +1,68 @@
...
+		return params.entrySet().stream().collect(Collectors
+				.toMap(entry -> WellKnownAuthParameters.getCamelCaseKey(entry.getKey()), Map.Entry::getValue));
+	}
+
+	private static Map<String, String> convertKebabCaseKeysToCamelCase(Map<String, String> params) {
+		return params.entrySet().stream()
+				.collect(Collectors.toMap(entry -> convertKebabCaseToCamelCase(entry.getKey()), Map.Entry::getValue));
+	}
+
+	protected static String maybeConvertToEncodedParamString(Map<String, String> params) {
```
onobc, line 55: **[NIT]** Class is public but the exposed method is protected, I would probably just make them both package protected for consistency.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
alpreu: Ah I see, good point. Yes, let's throw an exception if both are set

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @author Alexander Preuß
+ */
+public final class WellKnownAuthParameters {
+
+	private static final Map<String, String> lowerCaseToCamelCase = new HashMap<>();
+
+	static {
+		// Athenz
+		lowerCaseToCamelCase.put("tenantdomain", "tenantDomain");
```
onobc, line 33: I wonder if these are available in the auth modules? I'm not sure if I would want to pull them in as references even if they were available but it may be nice to just have a constant for the auth param (in camel case) then we could just use `toLowerCase` as the possibility for human error or having it in here 2x seems likely.  

Another option would be an enum for each that knows how to get its kebab case and lower case version. 

This is fine for now though. Just a couple of thoughts.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/AuthParameterUtils.java
```diff
@@ -0,0 +1,68 @@
...
+		if (params == null) {
+			return null;
+		}
+		params = convertWellKnownLowerCaseKeysToCamelCase(params);
```
onobc, line 59: It may be helpful to add a comment above this about the fact that env vars will come in as lower-case as we recommend mapping the keys as env vars as a single word w/no underscores and this is how we "reconstitute" them to be in their proper camel-case form.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AuthParameterUtilsTests.java
```diff
@@ -0,0 +1,56 @@
...
+
+	private static Stream<Arguments> encodedParamStringConversionProvider() {
+		return Stream.of(
+				arguments("camelCase",
+						Map.of("issuerUrl", "https://auth.server.cloud", "privateKey", "file://Users/xyz/key.json",
+								"audience", "urn:sn:pulsar:abc:xyz")),
+				arguments("kebabCase",
+						Map.of("issuer-url", "https://auth.server.cloud", "private-key", "file://Users/xyz/key.json",
+								"audience", "urn:sn:pulsar:abc:xyz")),
+				arguments("lowerCase", Map.of("issuerurl", "https://auth.server.cloud", "privatekey",
```
onobc, line 58: Should we add another case where there are camelCase, kebabCase, and lowerCase all in the same params map? Not sure if there is any value in doing so or not?

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AuthParameterUtilsTests.java
```diff
@@ -0,0 +1,62 @@
...
+		return Stream.of(arguments("null", null), arguments("empty", Collections.emptyMap()),
+				arguments("camelCase",
+						Map.of("issuerUrl", "https://auth.server.cloud", "privateKey", "file://Users/xyz/key.json",
+								"audience", "urn:sn:pulsar:abc:xyz")),
+				arguments("kebabCase",
+						Map.of("issuer-url", "https://auth.server.cloud", "private-key", "file://Users/xyz/key.json",
+								"audience", "urn:sn:pulsar:abc:xyz")),
+				arguments("lowerCase", Map.of("issuerurl", "https://auth.server.cloud", "privatekey",
+						"file://Users/xyz/key.json", "audience", "urn:sn:pulsar:abc:xyz")));
+	}
```
onobc, line 60: Wdy about adding a test that uses the mechanism I described [here](https://github.com/spring-projects-experimental/spring-pulsar/pull/97#issuecomment-1242223676) which would test the env var case end-end.  I think this would require SpringBootTest in order to do the relaxed env var bindings and that does not exist in this module. Would need to do this test in the autoconfigure module. This is fine for now - I may explore doing that later.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Returns the camel-cased version a Pulsar auth parameter or the given key in case it
+	 * is not part of the well-known ones.
+	 * @param lowerCaseKey the lower-cased auth parameter
+	 * @return the camel-cased auth parameter, or the lowerCaseKey if the parameter is not
+	 * found.
+	 */
+	public static String getCamelCaseKey(String lowerCaseKey) {
+		return lowerCaseToCamelCase.getOrDefault(lowerCaseKey, lowerCaseKey);
```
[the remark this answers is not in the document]
alpreu: I think I would keep it as is for now, the main benefit I see is that in case we get out of sync with the Pulsar repo it does not prevent the user from setting their keys (or also if they implement a custom Authentication)

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AuthParameterUtilsTests.java
alpreu: Sure, I'll add that

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
alpreu: >I wonder if these are available in the auth modules?
Some of them are, unfortunately not all of them...

You're right about the `toLowerCase`, I created a followup ticket (#111)

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/AuthParameterUtilsTests.java
alpreu: Sure, let's add it once your PR is merged so we can reuse the test class :)

onobc: I created https://github.com/spring-projects-experimental/spring-pulsar/issues/112 as a follow on

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/ee194f4a6d0f2abe7a778b0f09ad5b2d588abdb7 

See https://github.com/spring-projects-experimental/spring-pulsar/issues/95#issuecomment-1245920730 for follow on comment

Issue #111, opened 2022-09-13
Refactor WellKnownAuthParameters


alpreu:
Right now for every key in `WellKnownAuthParameter` we have to add both the lowercased as well as the camelCased version manually. @onobc had some ideas in how to improve this [here](https://github.com/spring-projects-experimental/spring-pulsar/pull/97#discussion_r969654713)

PR #116, opened 2022-09-14
Refactor: Define constants in WellKnownAuthParameters

alpreu:
This PR refactors the map in `WellKnownAuthParameters` to use constants for the individual parameters (#111)

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
@@ -80,7 +80,7 @@ enum WellKnownAuthParameters {
 
 	private static final Map<String, String> LOWER_CASE_TO_CAMEL_CASE = Arrays.stream(values())
 			.map(WellKnownAuthParameters::getCamelCaseKey)
-			.collect(Collectors.toMap(Function.identity(), WellKnownAuthParameters::toCamelCaseKey));
+			.collect(Collectors.toMap(String::toLowerCase, Function.identity()));
 
 	private final String camelCaseKey;
 
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
@@ -24,98 +24,70 @@ import java.util.Map;
  *
  * @author Alexander Preuß
  */
-public final class WellKnownAuthParameters {
+enum WellKnownAuthParameters {
+
+	TENANT_DOMAIN("tenantDomain"),
+
+	TENANT_SERVICE("tenantService"),
+
+	PROVIDER_DOMAIN("providerDomain"),
+
+	PRIVATE_KEY("privateKey"),
+
+	PRIVATE_KEY_PATH("privateKeyPath"),
+
+	KEY_ID("keyId"),
+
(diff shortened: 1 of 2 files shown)


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
```diff
@@ -28,40 +28,90 @@ public final class WellKnownAuthParameters {
 
 	private static final Map<String, String> lowerCaseToCamelCase = new HashMap<>();
 
+	private static final String TENANT_DOMAIN = "tenantDomain";
```
onobc, line 31: This is definitely an impovement @alpreu . If we made these an enum we could then also iterate through them to add their camel case mapping and avoid the N calls for each of these below. Wdyt?

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
```diff
@@ -24,48 +24,70 @@
...
-		lowerCaseToCamelCase.put("scope", "scope");
-		// Sasl
-		lowerCaseToCamelCase.put("sasljaasclientsectionname", "saslJaasClientSectionName");
-		lowerCaseToCamelCase.put("servertype", "serverType");
-		// Tls
-		lowerCaseToCamelCase.put("tlscertfile", "tlsCertFile");
-		lowerCaseToCamelCase.put("tlskeyfile", "tlsKeyFile");
-		// Token
-		lowerCaseToCamelCase.put("token", "token");
+		for (WellKnownAuthParameters param : values()) {
```
onobc, line 82: Ahhh yes! Big improvement. We can tweak it one step further w/ stream API such as: 
```java
private static final Map<String, String> LOWER_CASE_TO_CAMEL_CASE = Arrays.stream(values())
		.map(WellKnownAuthParameters::getCamelCaseKey)
		.collect(Collectors.toMap(Function.identity(), WellKnownAuthParameters::toCamelCaseKey));
```
NOTE: This would require renaming the existing static method to `toCamelCaseKey` as you would need a getter to get the camel case key. 

I always see how far I can get w/ the streams API and sometimes it turns out less readable etc.. I'm good w/ the code as-is also.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
```diff
@@ -24,48 +24,70 @@
...
+			lowerCaseToCamelCase.put(param.camelCaseKey.toLowerCase(), param.camelCaseKey);
+		}
 	}
 
-	private WellKnownAuthParameters() {
+	public final String camelCaseKey;
```
onobc, line 87: Can this be private (expose a getter instead?)

onobc (approved):

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/WellKnownAuthParameters.java
alpreu: That removes the ugly `static {...}` block completely, very nice!

onobc (commented):
Looks good @alpreu . Thanks for the updates. It looks like the tests are failing w/ latest changes though. Once we get that taken care of we can merge this.

alpreu:
> Looks good @alpreu . Thanks for the updates. It looks like the tests are failing w/ latest changes though. Once we get that taken care of we can merge this.

Oops, did not check before pushing. There was an issue with the mapping functions. Should be good now ;)

PR #118, opened 2022-09-16
Listener container properties for batch receive.

sobychacko:
Acknowledgment API cleanup.

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
@@ -1323,6 +1323,21 @@ public class PulsarProperties {
 		 */
 		private SchemaType schemaType;
 
+		/**
+		 * Max number of messages for batch.
+		 */
+		private int maxNumMessages = -1;
+
+		/**
+		 * Max number of bytes for batch.
+		 */
+		private int maxNumBytes = 10 * 1024 * 1024;
+
+		/**
+		 * Timeout for batch.
+		 */
+		private int batchTimeout = 100;
+
 		public AckMode getAckMode() {
 			return this.ackMode;
 		}
@@ -1337,6 +1352,30 @@ public class PulsarProperties {
 
 		public void setSchemaType(SchemaType schemaType) {
 			this.schemaType = schemaType;
+		}
+
+		public int getMaxNumMessages() {
+			return this.maxNumMessages;
+		}
+
+		public void setMaxNumMessages(int maxNumMessages) {
+			this.maxNumMessages = maxNumMessages;
+		}
+
+		public int getMaxNumBytes() {
+			return this.maxNumBytes;
+		}
+
+		public void setMaxNumBytes(int maxNumBytes) {
+			this.maxNumBytes = maxNumBytes;
+		}
+
+		public int getBatchTimeout() {
+			return this.batchTimeout;
+		}
+
+		public void setBatchTimeout(int batchTimeout) {
+			this.batchTimeout = batchTimeout;
 		}
 
 	}
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
@@ -360,6 +360,11 @@ public class DefaultPulsarMessageListenerContainer<T> extends AbstractPulsarMess
 									else if (this.containerProperties.getAckMode() == AckMode.BATCH) {
 										this.nackableMessages.add(message.getMessageId());
 									}
+									else {
+										throw new IllegalStateException(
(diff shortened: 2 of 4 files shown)


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1323,6 +1323,21 @@ public static class Listener {
...
+
+		/**
+		 * Max number of bytes for batch.
+		 */
+		private int maxNumBytes = 10 * 1024 * 1024;
+
+		/**
+		 * Timeout for batch.
+		 */
+		private int batchTimeout = 100;
```
onobc, line 1339: I think we should move this to a `Duration` because it is ambiguous as to what it is currently (millis, seconds?).

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1323,6 +1323,21 @@ public static class Listener {
...
+		 */
+		private int maxNumMessages = -1;
+
+		/**
+		 * Max number of bytes for batch.
+		 */
+		private int maxNumBytes = 10 * 1024 * 1024;
+
+		/**
+		 * Timeout for batch.
```
onobc, line 1337: It may be helpful to  elaborate on "for batch" (this applies to all 3 of the new fields)

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -360,6 +360,11 @@ else if (this.listener != null) {
 									else if (this.containerProperties.getAckMode() == AckMode.BATCH) {
 										this.nackableMessages.add(message.getMessageId());
 									}
+									else {
+										throw new IllegalStateException(
+												"Exception occurred and not negatively acknowledged or handled properly",
```
onobc, line 365: Is there any action we can recommend for the user to take in this case? I guess, what is the outcome in this case (data loss?)

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -541,12 +546,29 @@ public void acknowledge() {
 
 		@Override
 		public void acknowledge(MessageId messageId) {
-			throw new UnsupportedOperationException();
+			try {
```
onobc, line 549: Wdyt about 

1. moving this out to a method that gets used by here and the list version below.
2. log the exception message and state that the ack failed and that we are now instead nacking? If so, what level should that be at DEBUG/INFO/WARN? Will this happen often or is it rare?

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1323,6 +1323,21 @@ public static class Listener {
 		 */
 		private SchemaType schemaType;
 
+		/**
+		 * Max number of messages for batch.
+		 */
+		private int maxNumMessages = -1;
+
+		/**
+		 * Max number of bytes for batch.
```
onobc, line 1332: This could be a candidate for `DataSize`, although the unit is implied by the name currently.

onobc (changes requested):
@sobychacko I requested some minor changes around config props name/type/docs.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
onobc: https://github.com/spring-projects-experimental/spring-pulsar/issues/119 for later

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
onobc: https://github.com/spring-projects-experimental/spring-pulsar/issues/119

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
onobc: https://github.com/spring-projects-experimental/spring-pulsar/issues/120 turns out to be a bigger fish to fry than this section in this PR. We can handle this post M1.

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/eef1616aecfea9f01b1a7dce096b3ae5a53fc7c7

PR #122, opened 2022-09-16
Use enum for PulsarListener subscriptionType

onobc:
**NOTE** I also took the liberty of simplifying some `if/else/return` blocks and moving `.collect(Collectors.toList()) -> `toList()`.


sobychacko:
LGTM. Merging upstream.

PR #123, opened 2022-09-16
Batch receive properties not copied properly

sobychacko:
* When Boot provides properties for batch receive, they are not copied downstream through the ConcurrentPulsarListenerContainerFactory. Fixing this issue.

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
@@ -16,10 +16,14 @@
 
 package org.springframework.pulsar.config;
 
+import java.util.Arrays;
+
 import org.apache.commons.logging.LogFactory;
 import org.apache.pulsar.client.api.Schema;
 
+import org.springframework.beans.BeanWrapper;
 import org.springframework.beans.BeansException;
+import org.springframework.beans.PropertyAccessorFactory;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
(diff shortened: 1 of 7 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
```diff
@@ -171,11 +175,27 @@ else if (this.autoStartup != null) {
 			instance.setAutoStartup(this.autoStartup);
 		}
 
+		copyProperties(this.containerProperties, instance.getContainerProperties(),
```
onobc, line 178: I like the dynamic nature of this copy method, **but** it is inconsistent w/ the rest of the properties. It makes me wonder what is special about these 3 vs. the other 5 on line 181-185. For consistency I think it would be good to just fold these 3 properties into the `acceptIfNotNull` chain below. Wdyt?

sobychacko: ya, that is fine. We can do that.

spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
```diff
@@ -171,11 +175,27 @@ else if (this.autoStartup != null) {
 			instance.setAutoStartup(this.autoStartup);
```
onobc, line 175: I noticed when looking at the copy properties that we have this 3rd way commented out above ...

```java
		// BeanUtils.copyProperties(this.containerProperties, properties, "topics",
		// "messageListener",
		// "batchListener", "subscriptionName", "subscriptionType", "schema");
```
we should probably yank this out. I can see though why that approach would be nice - we would not miss newly added props. The downside is. you can't do things like "set only if not set already for these fields but always copy these fields".

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/685472b9966918ef707ac5a7f29285ecfe95ffb9 and https://github.com/spring-projects-experimental/spring-pulsar/commit/9d48b8df1b3987709fbc2cf230299f0ff4f12ed4

Issue #133, opened 2022-09-20
Improvements on the usage of ContainerProperties and retrying AtomicBoolean


sobychacko:
### Discussed in https://github.com/spring-projects-experimental/spring-pulsar/discussions/132

<div type='discussions-op-text'>

<sup>Originally posted by **tomazfernandes** September 20, 2022</sup>
Hey there! Congrats on the first Milestone!

I have a couple of quick suggestions for your consideration.

In these and a couple of other parts you lookup the `ContainerProperties` when the Container is already running. Unless I'm missing something, `ContainerProperties` is a mutable object and users can change those settings at will.

https://github.com/spring-projects-experimental/spring-pulsar/blob/10ef7000a9ab6c28768fa3aa53fb6431b06ea23f/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java#L285-L289

https://github.com/spring-projects-experimental/spring-pulsar/blob/10ef7000a9ab6c28768fa3aa53fb6431b06ea23f/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java#L300-L304

Is a user changing these settings with the Container running a feature? Otherwise, isn't it better to maybe extract these properties to `final` fields in the inner `Listener` class when the container starts and look that up instead?

Another part I'm not too sure about is this one:

https://github.com/spring-projects-experimental/spring-pulsar/blob/10ef7000a9ab6c28768fa3aa53fb6431b06ea23f/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java#L456-L458

You seem to be using a concurrency structure in a non concurrency-safe way. If this isn't concurrent code, why not use a plain boolean? Or, if it is concurrent code, isn't the `compareAndSet` method a better choice, instead of using separate `get` and `set` calls?

WDYT, makes sense, or maybe I'm missing something? Thanks!</div>

PR #140, opened 2022-09-21
Code cleanup in message listener container

sobychacko:
- More controlled access to ContainerProperties in the listener
- Concurrency related cleanup in retrying code

Resovles https://github.com/spring-projects-experimental/spring-pulsar/issues/133

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
@@ -170,6 +170,20 @@ public class DefaultPulsarMessageListenerContainer<T> extends AbstractPulsarMess
 
 		private final PulsarConsumerErrorHandler<T> pulsarConsumerErrorHandler;
 
+		private final boolean isBatchListener = this.containerProperties.isBatchListener();
+
+		private final boolean isManualAck = this.containerProperties.getAckMode().equals(AckMode.MANUAL);
+
+		private final boolean isBatchAck = this.containerProperties.getAckMode().equals(AckMode.BATCH);
+
+		private final boolean isRecordAck = this.containerProperties.getAckMode().equals(AckMode.RECORD);
+
+		private final boolean isSharedSubscription = this.containerProperties.getSubscriptionType()
+				.equals(SubscriptionType.Shared);
+
+		private final boolean isKeySharedSubscription = this.containerProperties.getSubscriptionType()
+				.equals(SubscriptionType.Key_Shared);
+
 		@SuppressWarnings({ "unchecked", "rawtypes" })
 		Listener(MessageListener<?> messageListener) {
 			if (messageListener instanceof PulsarBatchMessageListener) {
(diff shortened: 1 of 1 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -170,6 +170,20 @@ private final class Listener implements SchedulingAwareRunnable {
 
 		private final PulsarConsumerErrorHandler<T> pulsarConsumerErrorHandler;
 
+		private final boolean isBatchListener = this.containerProperties.isBatchListener();
+
+		private final boolean isManualAck = this.containerProperties.getAckMode().equals(AckMode.MANUAL);
```
onobc, line 175: I think holding onto every possible combination of field from the CP that we need is going to grow and make this harder to maintain and less readable. Wdyt instead of simply holding onto the piece of data rather than the query on the data? In the case of the `AckMode`, we would not have 3 fields but rather 1 for the AckMode. Then the code would read how it did before with `this.ackMode...` vs. `this.containerProperties.getAckMode()...`.

onobc (changes requested):

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -453,27 +461,21 @@ private void invokeRecordListenerErrorHandler(AtomicBoolean inRetryMode, Message
...
 
 		private void pendingMessagesHandledSuccessfully(AtomicBoolean inRetryMode,
 				AtomicBoolean messagesPendingInBatch) {
-			if (inRetryMode.get()) {
-				inRetryMode.set(false);
-			}
-			if (messagesPendingInBatch.get()) {
-				messagesPendingInBatch.set(false);
-			}
+			inRetryMode.compareAndSet(true, false);
```
tomazfernandes, line 477: This is just an implementation detail, so definitely no biggie. But on a second look, seems this code is not executed concurrently in any point, only sequentially inside the main `while` and `Message` loops. So unless I'm missing something a simple `boolean` might be simpler and avoid confusion.

sobychacko: The reason why we went with `AtomicBoolean` is so that we can modify it inside some lambda expressions. If it is primitive `boolean` we need to make it `final` to use inside the lambda, but then we cannot modify it. `AtomicBoolean` fixes that issue. You are right; this code is not accessed concurrently from anywhere else.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -172,11 +172,7 @@ private final class Listener implements SchedulingAwareRunnable {
...
 		private final boolean isBatchListener = this.containerProperties.isBatchListener();
 
-		private final boolean isManualAck = this.containerProperties.getAckMode().equals(AckMode.MANUAL);
-
-		private final boolean isBatchAck = this.containerProperties.getAckMode().equals(AckMode.BATCH);
-
-		private final boolean isRecordAck = this.containerProperties.getAckMode().equals(AckMode.RECORD);
+		private final AckMode ackMode = this.containerProperties.getAckMode();
 
 		private final boolean isSharedSubscription = this.containerProperties.getSubscriptionType()
```
onobc, line 177: I did not call it out explicitly, but for consistency do you mind doing the same for subscritpion type?

onobc (approved):

Issue #120, opened 2022-09-16
Consolidate ACK/NACK calls in DefaultPulsarMessageListenerContainer


onobc:
There is quite a bit of duplication in DPMLC around ack/nack. We always try to ack and if that fails we nack. 

1. Lets log when the ack fails (INFO/DEBUG/WARN level?) w/ message id for debugging purposes
2. Lets keep it DRY by consolidating to a comment method that does this ack-then-nack-if-the-ack-fails logic.
_Originally posted by @onobc in https://github.com/spring-projects-experimental/spring-pulsar/pull/118#discussion_r972581912_

PR #141, opened 2022-09-22
Consolidate ACK/NACK calls in container

sobychacko:
Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/120

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
@@ -32,6 +32,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import org.apache.commons.logging.LogFactory;
 import org.apache.pulsar.client.api.BatchReceivePolicy;
 import org.apache.pulsar.client.api.Consumer;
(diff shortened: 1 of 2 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -514,44 +516,34 @@ private void handleAcks(Messages<T> messages) {
...
-		public void acknowledge(MessageId messageId) {
+		private static void handleAckByMessageId(Consumer<?> consumer, MessageId messageId) {
 			try {
-				this.consumer.acknowledge(messageId);
+				consumer.acknowledge(messageId);
 			}
-			catch (PulsarClientException e) {
-				this.consumer.negativeAcknowledge(messageId);
+			catch (PulsarClientException pce) {
+				AbstractAcknowledgement.logger.warn(pce,
```
onobc, line 544: **[NIT]** Inconsistent qualifying as we use `AbstractAcknowledgement.logger` here but above on 536 we just call `handleAckByMessageId`. I am ok w/ or w/o the classname qualfier, as long as they are consistent

onobc: Going to handle in subsequent PR.

onobc (approved):

Issue #143, opened 2022-09-23
DnsResolverUtil errors when running apps


onobc:
When running the sample apps or during testing of the sanity app we see the following exception in the logs:
```java
java.lang.IllegalAccessException: class org.apache.pulsar.common.util.netty.DnsResolverUtil cannot access class sun.net.InetAddressCachePolicy (in module java.base) because module java.base does not export sun.net to unnamed module @3d82c5f3
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:392) ~[na:na]
```
which is related to Java17 and add-opens. 

Let's update this in the proper location to avoid this exception. Do we need to update this in the docs anywhere?

PR #145, opened 2022-09-24
Remove DNSResolverUtil error when running tests

onobc:
See #143

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/gradle/JavaConventionsPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/gradle/JavaConventionsPlugin.java
@@ -173,6 +173,11 @@ public class JavaConventionsPlugin implements Plugin<Project> {
 				testLoggingContainer.setShowStackTraces(true);
 				testLoggingContainer.setExceptionFormat(TestExceptionFormat.FULL);
 			});
+			test.jvmArgs(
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/java/org/springframework/boot/gradle/JavaConventionsPlugin.java
```diff
@@ -173,6 +173,11 @@ private void configureTestConventions(Project project) {
 				testLoggingContainer.setShowStackTraces(true);
 				testLoggingContainer.setExceptionFormat(TestExceptionFormat.FULL);
 			});
+			test.jvmArgs(
```
onobc, line 176: @sobychacko this removes the warning when running tests in any project. 

Where else were you seeing the warning? I thought it was showing up when running the sample apps but I am unable to reproduce that now (I wonder if the DNS is cached and then not hitting that code path that triggers it). I have a fix queued up for the sample apps if I can prove it is happening there.

sobychacko: @onobc I think I saw those warnings when running the sample apps.

onobc: @sobychacko  Can you see if you can reproduce it?

PR #146, opened 2022-09-27
Add Observations for send/receive

onobc:
This adds observations to `PulsarTemplate` and `DefaultPulsarMessageListenerContainer`.

### Remaining items (follow on PRs)
- [x] auto-generate docs and plug-in to newly added observability section in ref guide
- [x] ~add properties and hook into our Spring Boot starter where it makes sense~ (#147)

Big thanks to @artembilan and @garyrussell for trailblazing this in `spring-kafka` as this was my reference model for this proposal. 

@marcingrzejszczak @jonatan-ivanov if either of you get a chance to give this a once-over that would be great. However, I know that treatment was applied to `spring-kafka` and as stated above, this was modeled from that. I am fairly certain there will be no surprises here.


spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
```diff
@@ -118,7 +118,7 @@ public void afterPropertiesSet() {
 	@Override
 	public C createListenerContainer(PulsarListenerEndpoint endpoint) {
 		C instance = createContainerInstance(endpoint);
-		JavaUtils.INSTANCE.acceptIfNotNull(endpoint.getSubscriptionName(), instance::setBeanName);
+		JavaUtils.INSTANCE.acceptIfNotNull(endpoint.getId(), instance::setBeanName);
```
onobc, line 121: @sobychacko let's chat about this change. I am not sure if you were using the sub name as the id always has the child container index on it or some other reason.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -92,14 +115,52 @@ public SendMessageBuilder<T> newMessage(T message) {
...
+	 * Set a custom observation convention.
+	 * @param observationConvention the convention.
+	 */
+	public void setObservationConvention(PulsarTemplateObservationConvention observationConvention) {
+		this.observationConvention = observationConvention;
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		// TODO is this how we want to do this? What about SBAC?
```
onobc, line 154: Ahh yes.... the green and red of Github diff tool shows me things I forgot to remove before submitting. Will do that shortly.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -115,22 +176,49 @@ private CompletableFuture<MessageId> doSendAsync(String topic, T message,
...
-				// TODO success metrics
-			}
-			else {
-				this.logger.error(ex, () -> String.format("Failed to send msg to '%s' topic", topicName));
-				// TODO fail metrics
+
+		PulsarMessageSenderContext senderContext = PulsarMessageSenderContext.newContext(topicName, this.beanName);
+		Observation observation = newObservation(senderContext);
+		try {
+			observation.start();
```
onobc, line 183: I would like to clean this up a bit (here and in the listener counterpart) so that the try/observe/catch is around a singe method that "does the thing". As it is, the really decreases readability.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -341,37 +361,51 @@ public void run() {
...
-											? new ConsumerAcknowledgment(this.consumer, message) : null);
-								}
-								else if (this.listener != null) {
-									this.listener.received(this.consumer, message);
-								}
-								if (this.ackMode.equals(AckMode.RECORD)) {
-									handleAck(message);
-								}
-								inRetryMode.compareAndSet(true, false);
+							Observation observation;
```
onobc, line 364: **[NOTE]**: we are only observing single listener (aka record observations) currently. Same approach that spring-kafka has taken. We can look at how to add that to the batch case sensibly once the dust settles on this basic use case.

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Integration tests for {@link PulsarTemplateObservation send} and
+ * {@link PulsarListenerObservation receive} observations in Spring Pulsar against all
+ * supported Tracing runtimes.
+ *
+ * @author Chris Bono
+ * @see SampleTestRunner
+ */
+public class ObservationIntegrationTests extends SampleTestRunner {
```
marcingrzejszczak, line 65: Do you have a screenshot from Zipkin to see how this looks like?

marcingrzejszczak (approved):
I don't know the details of how Pulsar works, but from the point of view of Observability this looks good :) I would love to see a Zipkin screenshot to confirm the tracing part.

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
onobc: I have not done that yet but plan on doing so sometime today.  Yep, it is In the trace output (UI) is where we typically see where things are not all working end-end or need to be tweaked slightly.

onobc:
I went ahead and added the auto-generated docs as a 2nd commit in this PR. Looks good...

Screen Shot 2022-09-27 at 10 27 31 AM

Screen Shot 2022-09-27 at 10 28 41 AM

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -115,22 +176,49 @@ private CompletableFuture<MessageId> doSendAsync(String topic, T message,
...
-		return messageBuilder.sendAsync().whenComplete((msgId, ex) -> {
-			if (ex == null) {
-				this.logger.trace(() -> String.format("Sent msg to '%s' topic", topicName));
-				// TODO success metrics
-			}
-			else {
-				this.logger.error(ex, () -> String.format("Failed to send msg to '%s' topic", topicName));
-				// TODO fail metrics
+
+		PulsarMessageSenderContext senderContext = PulsarMessageSenderContext.newContext(topicName, this.beanName);
```
artembilan, line 180: You need to think about moving this object creation already after `Observation.NOOP`.
See `createNotStarted()` which accept `Supplier<T> contextSupplier` and does `NOOP` check.
You might not need that logic in your code afterwards.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -341,37 +361,51 @@ public void run() {
...
-							catch (Exception e) {
-								if (this.pulsarConsumerErrorHandler != null) {
-									invokeRecordListenerErrorHandler(inRetryMode, message, e);
-								}
-								else {
+							else {
+								observation = PulsarListenerObservation.LISTENER_OBSERVATION.observation(
+										this.containerProperties.getObservationConvention(),
+										DefaultPulsarListenerObservationConvention.INSTANCE,
+										new PulsarMessageReceiverContext(message, getBeanName()),
```
artembilan, line 372: Same have about `createNotStarted()` with a `Supplier` instead of this `if..else`

spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java
```diff
@@ -0,0 +1,88 @@
...
+		return this.destination;
+	}
+
+	/**
+	 * Acts as a carrier for a Pulsar message and records the propagated properties for
+	 * later access by the Pulsar message builder.
+	 */
+	public static class MessageHolder {
+
+		private Map<String, String> properties = new HashMap<>();
```
artembilan, line 76: `final`?

Also, I'm not sure that you'd to have this class instances be created outside of your API.
So, where is a `private` ctor?

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
```diff
@@ -0,0 +1,185 @@
...
+					.hasTimerWithNameAndTags("spring.pulsar.listener",
+							KeyValues.of("spring.pulsar.listener.id", "obs1-id-0"))
+					.hasTimerWithNameAndTags("spring.pulsar.listener",
+							KeyValues.of("spring.pulsar.listener.id", "obs2-id-0"));
+		};
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	@EnablePulsar
+	static class ObservationIntegrationTestAppConfig extends AbstractContainerBaseTests {
```
artembilan, line 113: Is it really intentional to extend from that class?

artembilan (changes requested):
Why generated docs are missed from the commit?

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
onobc: Yes as it ensures the PulsarTestContainer is up and running. I want to move that to a utils rather than have to subclass it everywhere.

spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java
onobc: I think we need to come up w/ consistent rules on when to use / not use final. It is mixed across projects currently. I think for cases where it is an instance variable and its immutable makes sense. I am not a fan of final everywhere which ultimately clutters the code. I just came from "that place" and I don't miss it. :) 

I believe Spring Boot takes this approach as well (not final everywhere).

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: Ah this would be nice. Thx. I will look into that here and below as well.

spring-pulsar/build.gradle
```diff
@@ -21,11 +22,17 @@ dependencies {
 	optional 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
 	optional 'com.fasterxml.jackson.datatype:jackson-datatype-joda'
 	optional 'com.jayway.jsonpath:json-path'
+	optional 'io.micrometer:micrometer-core'
+	optional 'io.micrometer:micrometer-tracing'
```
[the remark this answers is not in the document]
onobc: I will take another look - its possible they are not needed and I did not go back through to hygeine the deps.

spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java
artembilan: As far as I remember some Java compilers do an optimization for `final` properties: https://www.baeldung.com/java-final-performance

I'm not sure what you mean, though, about Spring Boot.
See, for example, its `AbstractNestedCondition`:
```
	private final ConfigurationPhase configurationPhase;

	AbstractNestedCondition(ConfigurationPhase configurationPhase) {
		Assert.notNull(configurationPhase, "ConfigurationPhase must not be null");
		this.configurationPhase = configurationPhase;
	}
```
or `PropertyMapper`:
```
	private final PropertyMapper parent;

	private final SourceOperator sourceOperator;

	private PropertyMapper(PropertyMapper parent, SourceOperator sourceOperator) {
		this.parent = parent;
		this.sourceOperator = sourceOperator;
	}
```

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: Unfortunately ...

1. Those built-in checks in `createNotStarted` will not take into account my local concerns for `containerProperties.isObservationEnabled `nor the null OR. In fact, the null OR will call the method to NPE as it does not tolerate a null OR being passed in.

2. The ObservationDocumentation.observe that takes a supplier only checks if registry.isNoop() before calling get() on the supplier which it then passes into Observeration.createNotStarted and does the other checks as:
```
if (registry == null || registry.isNoop()
        || !registry.observationConfig().isObservationEnabled(observationConvention.getName(), context)
        || observationConvention == NoopObservationConvention.INSTANCE) {
    return NOOP;
} else ...
```

For this to be useful in my case, I would need to be able to pass in some predicate to determine if it should be noop and also allow the supplier to make it further in the call chain before calling get().

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
onobc: Same as https://github.com/spring-projects-experimental/spring-pulsar/pull/146#discussion_r981645315

spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java
tomazfernandes: Just a quick 2c: AFAIK the criteria is we always add `final` to instance / class-level properties that are immutable, while local variables and method arguments we leave without (even if they are in fact final). This is important for enforcing immutable objects, thread-safety, etc.

For local variables I think compilers detect they don't mutate and apply any optimizations.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
artembilan: Sorry, I don't understand why you talk about `containerProperties.isObservationEnabled`.
You use `Observation.NOOP` condition in your own code. My point is to move that logic out of your code and supply context in lazy manner.
I believe that we step into this observation logic here only if that `containerProperties.isObservationEnabled == true`.
So, why do you try to add this variable into an equation where I just talked about deferring context creation down to the `createNotStarted()` call?

artembilan:
> Do you check them in SI  ?

I do at the moment. Plus I also added a condition to not run that Gradle task on CI.
My point was that this doc eventually going to be "very static" when the code base will have settled.

Although I'll debate with myself if this really has to be deferred to the docs creation phase instead.

spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
sobychacko: We discussed this internally with @onobc. We decided to use `id` as the unique way to discover listeners. So this change is acceptable.

onobc:
> > Do you check them in SI  ?
> 
> I do at the moment. Plus I also added a condition to not run that Gradle task on CI. My point was that this doc eventually going to be "very static" when the code base will have settled.
> 
> Although I'll debate with myself if this really has to be deferred to the docs creation phase instead.

That would be a true optimization (dont' run during build if its 99% static content). However, if the tool picks up changes (would not happen automatically unless version was set to `latest` or `snaphot` though) we would not see those w/o manually running. Also, people may forget to run the docs generator when something changes. They would have to know what those change-worthy things are.

spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java
onobc: @tomazfernandes that makes sense and sounds reasonable to me.

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
onobc: That is even better (interface). I will do this in follow on PR outside of this scope (see #148 )

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: I understand what you are wanting to do here, get rid of this if/else and let the createNotStarted+supplier handle this for us. I too don't like this if/else/create-myself pattern that we have to do. 

However, here is why the current impl of createNotStarted+supplier does not really help our cause. 

We have 2 custom conditions that dictate if we are using Observations or not, a boolean `isObservationEnabled` property and whether or not there is a OR in the AC. We have to respect these properties. None of the observation API calls will allow a null ObsReg either. 

Lets look at the suggested code path: 
```java
default <T extends Observation.Context> Observation observation(@Nullable ObservationConvention<T> customConvention,
        @NonNull ObservationConvention<T> defaultConvention, @NonNull Supplier<T> contextSupplier,
        @NonNull ObservationRegistry registry) {
    if (registry.isNoop()) {
        return Observation.NOOP;
    }
    return observation(customConvention, defaultConvention, contextSupplier.get(), registry);
}
```
From https://github.com/micrometer-metrics/micrometer/blob/13fbc3e6d6d8359920f4e99ab3408067f6e7b68f/micrometer-observation/src/main/java/io/micrometer/observation/docs/ObservationDocumentation.java#L209-L216

This is the API. you are suggesting to use, correct? It is the only one I see that accepts supplier. 


As you can see it only checks `registry.isNoop()` (it would NPE here already if we don't check the OR for null) and then it calls supplier.get() before sending to observation() method that in turn calls into Observation.createNotStarted which finally (after a few method chains) does the other checks such as: 

```java
static <T extends Context> Observation createNotStarted(ObservationConvention<T> observationConvention,
        @Nullable T context, @Nullable ObservationRegistry registry) {
    if (registry == null || registry.isNoop()
            || !registry.observationConfig().isObservationEnabled(observationConvention.getName(), context)
            || observationConvention == NoopObservationConvention.INSTANCE) {
        return NOOP;
    }
  return new SimpleObservation(observationConvention, registry, context == null ? new Context() : context);
}
```
From 
https://github.com/micrometer-metrics/micrometer/blob/13fbc3e6d6d8359920f4e99ab3408067f6e7b68f/micrometer-observation/src/main/java/io/micrometer/observation/Observation.java#L222-L230

 
1. I do not see a way to get rid of our `if/else` as we need to do our custom checks
2. The supplier.get is called immediately so I don't think it is a win either

Maybe I am missing something?

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
onobc: Resolving and will address (or not) based on comment above.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
artembilan: You do check for `OR` in your code, but you do not for `NOOP` and wrap your context object into `Supplier`.
That's how I see it for now.
Yes, Observation API might be improved with your investigation, but that's already outside of your `containerProperties.isObservationEnabled` condition.

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
onobc: @marcingrzejszczak I am actually going to do this when I do the SB starter side of this in #147 . I will ping you there when I do. Listing this comment in that issue so as to not forget - but resolving now as well.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: Got it. Yeh as it is we would not be able to get rid of the `if/else` and the `supplier` would buy little value as it is immediately invoked after only checking `registry.isNoop()`. It would need to be delayed until the actual `createNotStarted` to use that approach. 

I think I will leave it as-is based on my findings of the current API impl - unless you are strongly opposed :)

artembilan (approved):

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
onobc: Will handle/consider in #149

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: Will revisit in #147

artembilan (changes requested):
Did you see Checkstyle failures:
``` Error: eckstyle] [ERROR] /home/runner/work/spring-pulsar/spring-pulsar/spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java:31:1: Class PulsarMessageSenderContext should be declared as final. [FinalClass]
Error: eckstyle] [ERROR] /home/runner/work/spring-pulsar/spring-pulsar/spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java:74:9: Class MessageHolder should be declared as final. [FinalClass]
```
?

onobc:
> Did you see Checkstyle failures:

Yeh I did. I am going to squash/rebase a bit and will address then. 

> 
> ```
> Error: eckstyle] [ERROR] /home/runner/work/spring-pulsar/spring-pulsar/spring-pulsar/src/main/java/org/springframework/pulsar/observation/PulsarMessageSenderContext.java:74:9: Class MessageHolder should be declared as final. [FinalClass]
> ```
> 
>

sobychacko (approved):

Issue #150, opened 2022-09-28
Support Header annotations on PulsarListener


sobychacko:
Convert Pulsar Message metadata as `PulsarHeaders` and make them available on end-user applications on `PulsarListener` through Spring's `@Header` annotation.

PR #151, opened 2022-09-28
Support Pulsar Headers

sobychacko:
Convert Pulsar Message metadata as PulsarHeaders and make them available on PulsarListener through Spring's @Header annotation.

Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/150

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/support/PulsarHeaders.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/support/PulsarHeaders.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/support/PulsarHeaders.java
```diff
@@ -0,0 +1,116 @@
...
+
+	/**
+	 * The prefix for Pulsar headers.
+	 */
+	public static final String PREFIX = "pulsar_";
+
+	/**
+	 * The prefix for the message.
+	 */
+	public static final String PULSAR_MESSAGE = PREFIX + "message_";
```
onobc, line 34: Wdyt about shortening from `message_` to `msg_`?

sobychacko: Any particular reason for the abbreviation? I thought `messge` avoids any confusion.

PR #154, opened 2022-09-28
Use logback for tests (reduce logging)

onobc:
This PR sets logback as logging impl for tests in order to reduce test logging. 

### Points
- logback as test logging impl to respect `logback-test.xml`
- `logback-test.xml` to `WARN` by default 
- `testLogging.showStandardStreams = true` to see all `WARN` logs

### Results
All `WARN` logs are output to the console. This minimizes noise but allows us to see important warnings.


spring-pulsar/build.gradle
```diff
@@ -25,8 +25,7 @@ dependencies {
 	optional 'io.projectreactor:reactor-core'
 	testImplementation 'org.junit.jupiter:junit-jupiter'
 	testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
-	testRuntimeOnly 'org.apache.logging.log4j:log4j-core'
-	testRuntimeOnly 'org.apache.logging.log4j:log4j-jcl'
+	testRuntimeOnly("ch.qos.logback:logback-classic:1.4.1")
```
onobc, line 28: The tests were not respecting the logback-test.xml file as it was using log4j for tests. This falls in line w/ the default logging choice when using the SB starter (it uses logback by default as well).

PR #155, opened 2022-09-29
Use SpringBoot dependencies for version constraints

onobc:
### Main item
This PR replaces the many manual entries in the `spring-pulsar-dependencies` Gradle Java Platform.

All removed dependencies are covered by the SB BOM and were previously using the same exact version as in the SB BOM. We had to manually keep these in sync and it was toil. 
 
@sobychacko  This does not force SB on Spring Pulsar users as it is an internal project detail only.

### Piggyback commit
Further reduces the test logging output

### DIFF
--- a/spring-pulsar-dependencies/build.gradle
+++ b/spring-pulsar-dependencies/build.gradle
@@ -7,57 +7,24 @@ javaPlatform {
 }
 
 ext {
-	assertjVersion = '3.23.1'
-	awaitilityVersion = '4.2.0'
-	caffeineVersion = '3.1.1'
 	googleJsr305Version = '3.0.2'
-	hamcrestVersion = '2.2'
-	hibernateValidationVersion = '7.0.4.Final'
-	jacksonBomVersion = '2.13.3'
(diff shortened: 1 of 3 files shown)


spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/SpringPulsarBootAppSanityTests.java
```diff
@@ -34,7 +34,9 @@
  *
  * @author Chris Bono
  */
-@SpringBootTest(classes = SpringPulsarBootTestApp.class)
+@SpringBootTest(classes = SpringPulsarBootTestApp.class,
+		properties = { "spring.main.banner-mode=off"
```
onobc, line 38: The SB banner is logged very early, before the test log config file is respected apparently and it always was showing up in the output. This was the only test that was happening in and it is gone now. 

At this point the only thing we see in test output is WARN,ERROR logs.

sobychacko (approved):

PR #157, opened 2022-10-03
Pulsar header docs

sobychacko:
Resovles https://github.com/spring-projects-experimental/spring-pulsar/issues/153

Disable observation tests temporarily

### DIFF
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -193,7 +193,34 @@ public void listen3(Foo message) {
 Note the addition of a `schemaType` property on `PulsarListener`.
 That is because the library is not capable of inferring the schema type from the provided type `Foo`, we must tell the framework what schema to use.
 
-Here is an example of using `PulsarListener` to consume records in batches.
+Let us take a few more ways to consume single records using `PulsarListener` annotation.
+
+You can consume the Pulsar message directly as shown below:
+
+====
+[source, java]
+----
+@PulsarListener(subscriptionName = "my-subscription", topics = "my-topic")
+public void listen3(org.apache.pulsar.client.api.Message<String> message) {
+    System.out.println(message.getValue());
+}
+----
+====
+
+If you want to consume the record using the Spring messaging envelope, you can do it like this:
+
+====
+[source, java]
+----
+@PulsarListener(subscriptionName = "my-subscription-2", topics = "my-topic-2", schemaType = SchemaType.JSON)
+public void listen3(org.springframework.messaging.Message<String> message) {
+    System.out.println(message.getPayload());
+}
+----
+====
+
+Now let's see how we can consume records in batches.
+Here is an example of using `PulsarListener` to consume records in batches as POJO's.
 
 ====
 [source, java]
@@ -220,7 +247,20 @@ The following also should work in which we use the `Message` envelope provided b
 @PulsarListener(subscriptionName = "hello-batch-subscription", topics = "hello-batch", schemaType = SchemaType.JSON, batch = true)
 public void listen4(List<Message<Foo>> messages) {
     System.out.println("records received :" + messages.size());
-    messages.forEach((message) -> System.out.println("record : " + message));
+    messages.forEach((message) -> System.out.println("record : " + message.getValue()));
+}
+----
+====
+
+Here is another example of consuming batch records with the envelope of Spring messaging Message type.
+
+====
+[source, java]
+----
+@PulsarListener(subscriptionName = "hello-batch-subscription", topics = "hello-batch", schemaType = SchemaType.JSON, batch = true)
+public void listen4(List<org.springframework.messaging.Message<Foo>> messages) {
+    System.out.println("records received :" + messages.size());
+    messages.forEach((message) -> System.out.println("record : " + message.getPayload()));
 }
 ----
 ====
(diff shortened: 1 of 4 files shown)


spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -193,7 +193,34 @@ public void listen3(Foo message) {
 Note the addition of a `schemaType` property on `PulsarListener`.
 That is because the library is not capable of inferring the schema type from the provided type `Foo`, we must tell the framework what schema to use.
 
-Here is an example of using `PulsarListener` to consume records in batches.
+Let us take a few more ways to consume single records using `PulsarListener` annotation.
```
onobc, line 196: `Let us look at a few more ways` ?

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -193,7 +193,34 @@ public void listen3(Foo message) {
...
+}
+----
+====
+
+If you want to consume the record using the Spring messaging envelope, you can do it like this:
+
+====
+[source, java]
+----
+@PulsarListener(subscriptionName = "my-subscription-2", topics = "my-topic-2", schemaType = SchemaType.JSON)
```
onobc, line 215: Schema type and payload type do not match

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -193,7 +193,34 @@ public void listen3(Foo message) {
...
-Here is an example of using `PulsarListener` to consume records in batches.
+Let us take a few more ways to consume single records using `PulsarListener` annotation.
+
+You can consume the Pulsar message directly as shown below:
+
+====
+[source, java]
+----
+@PulsarListener(subscriptionName = "my-subscription", topics = "my-topic")
+public void listen3(org.apache.pulsar.client.api.Message<String> message) {
```
onobc, line 204: Is `listen3` copied from a test possibly? Can we drop the `3`?

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -220,7 +247,20 @@ The following also should work in which we use the `Message` envelope provided b
...
+----
+====
+
+Here is another example of consuming batch records with the envelope of Spring messaging Message type.
+
+====
+[source, java]
+----
+@PulsarListener(subscriptionName = "hello-batch-subscription", topics = "hello-batch", schemaType = SchemaType.JSON, batch = true)
+public void listen4(List<org.springframework.messaging.Message<Foo>> messages) {
```
onobc, line 261: Can we also do it w/ `Messages<Foo>`?

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -1222,6 +1262,116 @@ PulsarTopic partitionedTopic {
 ----
 ====
 
+[[pulsar-headers]]
+==== Pulsar Headers
+
+In this section, we will see how we can use the message metadata that Pulsar provides in its Message as Pulsar headers in a Spring application.
```
onobc, line 1268: I think we can simplify this w/ something like 
```java
... we can use the Pulsar message metadata as Pulsar headers...
```

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -1222,6 +1262,116 @@ PulsarTopic partitionedTopic {
...
+@PulsarListener(topics = "simpleListenerWithHeaders")
+void simpleListenerWithHeaders(String data, @Header(PulsarHeaders.MESSAGE_ID) MessageId messageId,
+                @Header(PulsarHeaders.RAW_DATA) byte[] rawData,
+                @Header("foo") String foo) {
+
+}
+----
+====
+
+In this example, we are consuming the data as `String`.
```
onobc, line 1293: It may be helpful to

1. say "In the above example..." as it may not be clear if it is talking about the next example. 
2. the fact that the data is string is probably not pertinent to the header infos
3. I think we can simplify this a bit. 

Maybe something like: 
```java
In the above example we are accessing the values for the `messageId` and `rawData` message metadata as well as a custom message property named "foo". The  Spring `@Header` annotation is used for each header field.
```

Wdyt?

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -1222,6 +1262,116 @@ PulsarTopic partitionedTopic {
...
+In addition, we are also extracting the message meatadata as Spring headers using the `Header` annotation from the Spring Framework.
+We just randomly picked the headers for message id and raw data.
+You can also extract custom message properties set when publishing using the `Header` annotation as observed in the retrieval of the property `foo` in the above example code.
+
+Now let us try to receive this same data, but with the envelope of a Pulsar `Message`.
```
onobc, line 1298: I would be in favor in reducing the code samples from 6 down to 2. Here is my reasoning:

The method of accessing the headers is the same in all cases (via the `@Header` annotation) w/ only 2 interesting facts:
1. When accessing in batch mode (aka N records) then there is a. parallel list of N header entries.
2. When accessing data as pulsar msg then you have the option of directly accessing the message metadata from the pulsar message

We could probably get by w/ 1 example and just talk to the points or at most 2 would cover the interesting pieces. This will simplify the maintenance of this doc and make it less likely to get bit rot. 

Wdyt?

spring-pulsar/src/main/java/org/springframework/pulsar/support/PulsarHeaders.java
```diff
@@ -34,82 +34,82 @@ public abstract class PulsarHeaders {
...
 
 	/**
-	 * Prefix for the raw message data.
+	 * Raw message data.
 	 */
 	public static final String RAW_DATA = PULSAR_MESSAGE + "raw_data";
 
 	/**
-	 * Prefix for message size.
+	 * Message size.
```
onobc, line 47: in what unit?

spring-pulsar/src/main/java/org/springframework/pulsar/support/PulsarHeaders.java
```diff
@@ -34,82 +34,82 @@ public abstract class PulsarHeaders {
...
 	 */
 	public static final String MESSAGE_SIZE = PULSAR_MESSAGE + "size";
 
 	/**
-	 * Prefix for message publish time.
+	 * Message publish time.
```
onobc, line 52: Any particular format?

onobc (changes requested):
@sobychacko looks good. A few comments to address - nothing major.

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
sobychacko: This is great feedback. Let's do this.

spring-pulsar/src/main/java/org/springframework/pulsar/support/PulsarHeaders.java
sobychacko: bytes. will updte.

onobc (approved):
Looks great @sobychacko . Thx for the updates.

PR #158, opened 2022-10-05
Add observations auto-configuration

### LINKED ISSUE #147: Support observations in Spring Boot starter
Once observations are in the core for send/receive let's make it easier for end users to consume.

### Easily disable/enable
- [ ] Add property to en/disable observations for send, receive, globally

### Easier customization
- [ ] Make it easier for users to add custom tags. As it stands they would need to code up a convention and set it on the observed target. Let's at least create a convention customizer that can be auto-configured. NOTE: SB auto-configures observation registry and other components. Can we leverage this in any way?

### Docs
- [ ] Make sure there is an easy to follow section in docs telling how end user can use this.
- [ ] Make sure there is an easy to follow section in docs telling how end user can access the built-in Pulsar consumer/producer stats.
### Random cleanup
- [ ] Revisit the observation if/else blockage in the send/receive methods to clean it up possibly

### Verify
- [ ] Test drive the above docs manually w/ Wavefront (or other platform). Ping @marcingrzejszczak when we do as he requested some screenshots in https://github.com/spring-projects-experimental/spring-pulsar/pull/146#discussion_r981315309.

onobc:
Adds auto-configuration of the observation aspects of the listener and template. 

### Points of interest
* I refactored the template and listeners to use immutable constructors for some key pieces to simplify them as well as make them auto-configuration friendly (eg. the auto-config will/not pass in the observation registry when user has disabled the obeservations property).

Resolves #147


sobychacko:
We did an internal code review by walking through the code. Looks good @onobc. Merged upstream.

PR #160, opened 2022-10-12
Add Micrometer Tracing to sample apps

onobc:
Also documents how to configure Micrometer Observations.

See #147


spring-pulsar-sample-apps/build.gradle
```diff
@@ -8,6 +8,12 @@ description = 'Spring Pulsar Sample Applications'
 dependencies {
 	api project(':spring-pulsar-spring-boot-starter')
 	implementation 'com.google.code.findbugs:jsr305'
+
+	// observability
```
onobc, line 12: I am going to add a section in the sample apps readme as well that shows how to run Zipkin locally. I may pull some of these doc details from the main docs to the sample app readme as they are a bit detailed currently.

Issue #161, opened 2022-10-12
Consumer restart causes duplicate messages


sobychacko:
Scenario:

1. Start an exclusive consumer--> Receive message.
2. Stop the consumer.
3. Restart the consumer or start another consumer under the same subscription name.
4. Occasionally, the same message that was already seen by the first consumer reappears in the second consumer.

The reason is because in step 2 above, when we stop the container if there are any `acks` in progress, they are immediately aborted by stopping the consumer and then the container is stopped. We need to wait for the consumer thread to finish what it is doing before stopping the consumer.

PR #162, opened 2022-10-12
Consumer restart causes duplicate messages

sobychacko:
When we stop the consumer, there is a race condition that forces the messages to not get acknowledged. This results in duplicate messages on the restart of the consumer.

This PR addresses the issue by waiting for the consumer thread to complete before stopping the consumer.

Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/161


spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -133,6 +137,15 @@ public void doStop() {
 		setRunning(false);
 		this.logger.info("Pausing this consumer.");
 		this.listenerConsumer.consumer.pause();
+		if (this.receiveInProgress) {
```
onobc, line 140: What is the implication of this interrupt if the batch receive is in the middle of a batch of records? I understand the need for this when the timeout is high - but it also feels like maybe this would be an operation we did when the user asked to "stopImmediately" vs. "stopGracefully".  Does this need to be included in this proposal or added when if we add a stopImmediately/Gracefully feature? 

Either way, a comment would help understand the purpose of the interrupt.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -133,6 +137,15 @@ public void doStop() {
...
+			this.listenerConsumerThread.interrupt();
+		}
+		try {
+			this.listenerConsumerThread.join();
+		}
+		catch (InterruptedException e) {
+			Thread.currentThread().interrupt();
```
onobc, line 147: May want to consider logging here as this is probably not a normal condition.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -286,6 +299,7 @@ public boolean isLongLived() {
 
 		@Override
 		public void run() {
+			DefaultPulsarMessageListenerContainer.this.listenerConsumerThread = Thread.currentThread();
```
onobc, line 302: Did checkstyle want the `OUTER.this` prefix? I would be in favor of dropping it if checkstyle does not complain.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -296,71 +310,83 @@ public void run() {
 				// Always receive messages in batch mode.
 				try {
 					if (!inRetryMode.get() && !messagesPendingInBatch.get()) {
+						DefaultPulsarMessageListenerContainer.this.receiveInProgress = true;
 						messages = this.consumer.batchReceive();
+						DefaultPulsarMessageListenerContainer.this.receiveInProgress = false;
```
onobc, line 315: If `batchReceive` fails for some other reason we do not clear the flag. I think we should move the set(false) to finally block rather than directly below the `batchReceive`.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -296,71 +310,83 @@ public void run() {
...
 				catch (PulsarClientException e) {
-					DefaultPulsarMessageListenerContainer.this.logger.error(e, () -> "Error receiving messages.");
-				}
-				Assert.isTrue(messages != null, "Messages cannot be null.");
-				if (this.isBatchListener) {
-					if (!inRetryMode.get() && !messagesPendingInBatch.get()) {
-						messageList = new ArrayList<>();
-						messages.forEach(messageList::add);
+					if (e.getCause() instanceof InterruptedException) {
+						DefaultPulsarMessageListenerContainer.this.logger.debug(e, () -> "Error receiving messages.");
```
onobc, line 320: I'm torn as to what level of logging we should use in this case. Would we want to ever see this message? How often would this be happening and be an expected thing? 

Regardless of level, I think we should differentiate this from the other error case w/ a more descriptive message about what is going on.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -296,71 +310,83 @@ public void run() {
...
-								try {
-									if (isSharedSubscriptionType()) {
-										this.consumer.acknowledge(messages);
+					else {
+						DefaultPulsarMessageListenerContainer.this.logger.error(e, () -> "Error receiving messages.");
+					}
+					messages = null;
+				}
+
+				if (messages != null) {
```
onobc, line 328: For readability, I would be in favor of inverting this and short circuiting (continue the loop) rather than indent the large block of logic. Something like 

```java
if (messages == null) {
  continue;
}

if (this.isBatchListener) {
  ...
else {
  ...
}
```
Wdyt?

spring-pulsar/src/test/java/org/springframework/pulsar/core/ConsumerAcknowledgmentTests.java
```diff
@@ -361,4 +361,64 @@ void testBatchNackForEntireBatchWhenUsingBatchListener() throws Exception {
 		pulsarClient.close();
 	}
 
+	// *********************
```
onobc, line 364: I'm not sure what the significance of this divider bar is, but do we need it in here?

spring-pulsar/src/test/java/org/springframework/pulsar/core/ConsumerAcknowledgmentTests.java
```diff
@@ -361,4 +361,64 @@ void testBatchNackForEntireBatchWhenUsingBatchListener() throws Exception {
...
+		pulsarContainerProperties.setMessageListener((PulsarRecordMessageListener<?>) (consumer, msg) -> {
+			counter1.getAndIncrement();
+		});
+		pulsarContainerProperties.setSchema(Schema.STRING);
+		DefaultPulsarMessageListenerContainer<String> container1 = new DefaultPulsarMessageListenerContainer<>(
+				pulsarConsumerFactory, pulsarContainerProperties);
+		container1.start();
+
+		Map<String, Object> prodConfig = new HashMap<>();
+		prodConfig.put("topicName", "duplicate-message-test");
```
onobc, line 387: Minor improvement would be `Collections.singletonMap`.

spring-pulsar/src/test/java/org/springframework/pulsar/observation/ObservationIntegrationTests.java
```diff
@@ -93,14 +93,6 @@ public SampleTestRunnerConsumer yourCode() {
...
 						"Message %s not received in listen2 (latchesByMessageListen1 = %s and latchesByMessageListen2 = %s)",
 						msg, listeners.latchesByMessageListen1, listeners.latchesByMessageListen2).isTrue();
-
-				// Without this sleep, the 2nd tracingSetup run sometimes fails due to
-				// messages from 1st run being
-				// delivered during the 2nd run. The test runs share the same listener
-				// config, including the
-				// same subscription names. Seems like the listener in run2 is getting
-				// message from run1.
-				Thread.sleep(5000);
```
onobc, line 103: Thank you for remembering my "workaround" ;)

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
```diff
@@ -262,6 +262,7 @@ void deadLetterPolicy() throws Exception {
 		// Normal consumer should receive 5 msg + 1 re-delivery
 		assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
 		container.stop();
+		dlqContainer.stop();
```
onobc, line 265: So this reminders me, in all of the consumer tests we close/stop/cleanup at the end of a **successful** test - but not during a failed one. 

I don't think we need to fix that in this PR, but do you mind adding an issue for us to cleanup properly so we don't forget to come back to this?

spring-pulsar/src/test/java/org/springframework/pulsar/core/ConsumerAcknowledgmentTests.java
```diff
@@ -361,4 +361,64 @@ void testBatchNackForEntireBatchWhenUsingBatchListener() throws Exception {
...
+				pulsarConsumerFactory, pulsarContainerProperties);
+		container2.start();
+
+		pulsarTemplate.send("hello john doe");
+
+		while (counter2.get() == 0) {
+			// busy wait until counter2 > 0
+		}
+		// Asserting that both consumers are only receiving the expected data.
+		assertThat(counter1.get()).isEqualTo(1);
```
onobc, line 417: I am sure you did this already, but this test fails w/o your fix, correct?

onobc (changes requested):
Nice fix @sobychacko ! I have left some minor suggestions / questions.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -74,14 +74,18 @@
...
 
 	private volatile Listener listenerConsumer;
 
 	private volatile CountDownLatch startLatch = new CountDownLatch(1);
 
 	private final AbstractPulsarMessageListenerContainer<?> thisOrParentContainer;
 
+	private volatile Thread listenerConsumerThread;
+
+	private volatile boolean receiveInProgress;
```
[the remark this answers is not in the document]
sobychacko: Yes, I agree. I just followed the same pattern that we use in a few other projects, but your suggestion is an improvement.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Yes, it was the checkstyle.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Ya, we can add that in a separate PR. I will add a comment.

sobychacko: https://github.com/spring-projects-experimental/spring-pulsar/issues/167

spring-pulsar/src/test/java/org/springframework/pulsar/core/ConsumerAcknowledgmentTests.java
sobychacko: lol - user error.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
sobychacko: https://github.com/spring-projects-experimental/spring-pulsar/issues/168

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Ok, I will clarify the log message. I was also struggling to decide what level should we log. We don't want this message to appear in the logs causing confusion to the user, but should be available for debugging purposes, thus went with `debug`.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Good idea.

onobc (approved):
Nice work @sobychacko - thx for the updates.

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/2360a9d2d1e1b5d16e526e7bfda14cc5474b4dce

PR #164, opened 2022-10-17
Add ReactivePulsarSenderTemplate

cbornet:


cbornet:
cc @onobc

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -0,0 +1,221 @@
...
+			observation.stop();
+			throw ex;
+		}
+	}
+
+	private Observation newObservation(PulsarMessageSenderContext senderContext) {
+		if (this.observationRegistry == null) {
+			return Observation.NOOP;
+		}
+		return PulsarTemplateObservation.TEMPLATE_OBSERVATION.observation(this.observationConvention,
```
cbornet, line 157: Is it OK to reuse PulsarTemplateObservation ?
Or should the reactive template have its own ?

cbornet: As far as the bean names are concerned there should be no collision.
For the configuration, it would be under `spring.pulsar.template`. We currently don't have anything else that could differ between regular and reactive template config, but maybe one day we could ?

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/DefaultReactivePulsarSenderFactory.java
```diff
@@ -0,0 +1,99 @@
...
+public class DefaultReactivePulsarSenderFactory<T> implements ReactivePulsarSenderFactory<T> {
+
+	private final LogAccessor logger = new LogAccessor(LogFactory.getLog(this.getClass()));
+
+	private final ReactivePulsarClient reactivePulsarClient;
+
+	private final ReactiveMessageSenderSpec reactiveMessageSenderSpec;
+
+	public DefaultReactivePulsarSenderFactory(PulsarClient pulsarClient,
+			ReactiveMessageSenderSpec reactiveMessageSenderSpec) {
```
onobc, line 49: The `ReactiveMessageSenderSpec` is akin to the map of config props in the non-reactive producer factor. And the SB auto-config will ultimately build that up and pass it in, correct?

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/DefaultReactivePulsarSenderFactory.java
```diff
@@ -0,0 +1,99 @@
...
+			sender.applySpec(this.reactiveMessageSenderSpec);
+		}
+		sender.topic(resolvedTopic);
+		if (messageRouter != null) {
+			sender.messageRouter(messageRouter);
+		}
+		if (!CollectionUtils.isEmpty(customizers)) {
+			customizers.forEach((c) -> c.customize(sender));
+		}
+		return sender.build();
```
onobc, line 91: One thing I noticed is there are no producer interceptors in the RMS. Has that not been added or is it achieved another way?

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderOperations.java
```diff
@@ -0,0 +1,102 @@
...
+	 */
+	ReactiveSendMessageBuilder<T> newMessage(T message);
+
+	/**
+	 * Builder that can be used to configure and send a message. Provides more options
+	 * than the send methods provided by {@link ReactivePulsarSenderOperations}.
+	 *
+	 * @param <T> the message payload type
+	 */
+	interface ReactiveSendMessageBuilder<T> {
```
onobc, line 62: I was going to save this comment as overall for the PR but I think it fits well here. 

The Reactive client has the concept of ReactiveMessageSender / ReactiveMessageSenderBuilder and MessageSpec / MessageSpecBuilder. When combined w/ the spring-pulsar concepts of ReactiveSendMessageBuilder and ReactivePulsarSenderFactory keeping track of whats-what was a bit confusing (for me) **at first**.  However, after digesting a bit, I think this is just the nature of things and the names you have chosen all make sense and line up w/ what was done for the non-reactive counterparts. 

That being said, I wonder if removing the `Reactive` prefix from `ReactiveSendMessageBuilder` -> `SendMessageBuilder` would simplify things a bit. I think we are safe from name collisions via packages. Wdyt?

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
onobc: > Is it OK to reuse PulsarTemplateObservation ?
> Or should the reactive template have its own Observation class ?

Excellent questions. I think using the same names w/ a distinguishing tag would work well. 

I would not imagine a single app instance using both reactive/non-reactive, but I could imagine a case where someone is doing a bake-off between the 2 and has 50% of their app instances using reactive and the other 50% non-reactive. It would be helpful be able to filter the metrics by client type etc..

All this being said, I would be in favor of pulling the O11Y out of the reactive side **for now**. The reason is it is a very new API and we may run into some kinks w/ it and I would prefer to get some runtime on it (and fix it in) in a single area before we proliferate it through the codebase. Wdyt?

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -0,0 +1,221 @@
...
+	private Mono<MessageId> doSend(String topic, T message,
+			MessageSpecBuilderCustomizer<T> messageSpecBuilderCustomizer, MessageRouter messageRouter,
+			ReactiveMessageSenderBuilderCustomizer<T> customizer) {
+		final String topicName = ReactiveMessageSenderUtils.resolveTopicName(topic, this.reactiveMessageSenderFactory);
+		this.logger.trace(() -> String.format("Sending reative msg to '%s' topic", topicName));
+
+		PulsarMessageSenderContext senderContext = PulsarMessageSenderContext.newContext(topicName, this.beanName);
+		Observation observation = newObservation(senderContext);
+		try {
+			observation.start();
```
onobc, line 124: If we did leave this in (see below comment) I wonder if the start call should go in the `doSubscribe` callback rather than outside the reactive chain. I would want to see what the Spring Framework is doing for WebClient observation instrumentation.

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageSenderFactoryTests.java
```diff
@@ -0,0 +1,159 @@
...
+				Arrays.asList(customizer1, customizer2));
+		sender.sendMessage(Mono.just(MessageSpec.of("test"))).block(Duration.ofSeconds(5));
+		assertSenderHasTopicAndRouter("topic1", router);
+	}
+
+	@Test
+	void createProducerWithNoTopic() {
+		ReactivePulsarSenderFactory<String> senderFactory = new DefaultReactivePulsarSenderFactory<>(pulsarClient,
+				null);
+		assertThatThrownBy(() -> senderFactory.createReactiveMessageSender(null, schema))
```
onobc, line 144: **[TinyNit / Tip]** I just learned this more specific AssertJ API (you can tell I just learned it because I have not used it anywhere in this project yet) but... check out 
```java
assertThatIllegalArgumentException()
		.isThrownBy(() -> senderFactory.createReactiveMessageSender(null, schema))
		.withMessageContaining("Topic must be specified when no default topic is configured");
```

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageSenderFactoryTests.java
```diff
@@ -0,0 +1,159 @@
...
+		when(producerBuilder.createAsync()).thenReturn(CompletableFuture.completedFuture(producer));
+		when(pulsarClient.newProducer(schema)).thenReturn(producerBuilder);
+	}
+
+	@Test
+	void createProducerWithSpecificTopic() {
+		ReactivePulsarSenderFactory<String> senderFactory = new DefaultReactivePulsarSenderFactory<>(pulsarClient,
+				null);
+		ReactiveMessageSender<String> sender = senderFactory.createReactiveMessageSender("topic1", schema);
+		sender.sendMessage(Mono.just(MessageSpec.of("test"))).block(Duration.ofSeconds(5));
```
onobc, line 79: I would consider using `StepVerifier` instead of the block.

onobc (changes requested):
Nice work @cbornet ! I think this is a great place to start. 

I have left some minor comments/suggestions but nothing major.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: OK. I can remove the O11y.
Then the value of the Spring template compared to the raw client seems very thin...

build.gradle
```diff
@@ -75,4 +75,11 @@ subprojects { subproject ->
 			}
 		}
 	}
+
+	repositories {
```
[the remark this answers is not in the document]
cbornet: OK. In the end I think we won't merge until we have a released version of the reactive client anyway.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactiveMessageSenderUtils.java
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	static <T> String resolveTopicName(String userSpecifiedTopic,
+			ReactivePulsarSenderFactory<T> reactiveMessageSenderFactory) {
+		ReactiveMessageSenderSpec reactiveMessageSenderSpec = reactiveMessageSenderFactory
+				.getReactiveMessageSenderSpec();
+		if (StringUtils.hasText(userSpecifiedTopic)) {
+			return userSpecifiedTopic;
+		}
+		else if (reactiveMessageSenderSpec != null && reactiveMessageSenderSpec.getTopicName() != null) {
```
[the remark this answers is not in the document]
cbornet: I've chosen this way because there is controversy on the usage of `Optional` to replace if/then/else (the original author of Optional advocating against it). One of the reason being wrapping is less performant.
But I can use the Optional if you prefer.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/DefaultReactivePulsarSenderFactory.java
cbornet: yes

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageSenderFactoryTests.java
cbornet: 👍

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageSenderFactoryTests.java
cbornet: Yes I was also considering it. Will take a look.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderOperations.java
cbornet: Sure, I'll rename it.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: done

build.gradle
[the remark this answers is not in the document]
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/DefaultReactivePulsarSenderFactory.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Default implementation of {@link ReactivePulsarSenderFactory}.
+ *
+ * @param <T> reactive sender type.
+ * @author Christophe Bornet
+ */
+public class DefaultReactivePulsarSenderFactory<T> implements ReactivePulsarSenderFactory<T> {
+
+	private final LogAccessor logger = new LogAccessor(LogFactory.getLog(this.getClass()));
```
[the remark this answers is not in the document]
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactiveMessageSenderUtils.java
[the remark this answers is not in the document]
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderOperations.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: Observations has been removed from now.
Will wait till it's hardened on the imperative side.

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageSenderFactoryTests.java
cbornet: done

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageSenderFactoryTests.java
onobc: Fair enough. Thx for looking into it.

onobc (approved):
Looks good @cbornet . Thx for the updates.

PR #165, opened 2022-10-19
[DO_NOT_MERGE_ILLUSTRATION_ONLY] Reactive module layout

onobc:
@cbornet So I dug around Spring Boot starters/autoconfigures and most of the modules do an additive reactive autoconfig where you get all of the imperative beans + reactive beans . Although, there are a few that completely split out the reactive and imperative autoconfig. 

This PR shows what it would look like if we split out the base modules common+imperative+reactive (note imperative is `spring-pulsar`). Described as follows:

- `spring-pulsar-common` new module containing admin + client used by both imperative + reactive
- `spring-pulsar-reactive` new module containing only reactive aspects
- `spring-pulsar` depends on `spring-pulsar-common`
- `spring-pulsar-reactive` depends on `spring-pulsar-common`

- `spring-pulsar-spring-boot-autoconfigure` configures admin + client and optionally imperative + reactive (if they are on classpath)
- `spring-pulsar-spring-boot-starter` depends on spsba and puts` spring-pulsar` on classpath
- `spring-pulsar-reactive-spring-boot-starter` depends on spsba and puts `spring-pulsar-reactive` on classpath

Here are my current thoughts:
- Until we get the Reactive consumer side of things fleshed out a bit it may be premature to split the modules out. I think it will reveal some more common components and may influence the split.

- What I would like to do is park this PR (namely my commit - the third commit) until we get a bit further and then decide how to break things out. It will quite a bit of distracting work to do and I would rather do it once than a few times as we go. In the meantime I think we should do something that looks more like this subset PR (https://github.com/spring-projects-experimental/spring-pulsar/pull/166/commits/e611c0a5b0a5bc6ff3f3a2982f4ca4720af0a1e7).  

**QUESTION**: I know we want to allow users to say "I don't want anything reactive". However, do we want to allow them to say "I only want things reactive - leave out the imperative stuff"?


migroskub:
@onobc users should not really have a concern about wishing to merge reactive with non-reactive on the same application. This may happen when migrating a synchronous application to an async one, but I don't see a use case of using both reactive and non-reactive for the same API. Sounds quite weird.

migroskub:
@onobc and btw (maybe not exactly related to this issue specifically) - maybe we should auto-configure a list of `Reactive`/`PulsarProperties` instead of `Reactive`/`PulsarProperties`? This way it will be possible to write/ read with totally different configurations.

onobc:
I ❤️ this guy! Thank you @migroskub for keeping a close eye on things. 

And yep, that is exactly the plan. This PR was just me noodling a couple of different approaches and getting it into a presentable (not perfect) state. Me and @cbornet synced up on this today and when we did I pointed out that the `PulsarProperties` is meant to be `ReactivePulsarProperties`.  We can harmonize them later if it makes sense to do so. 

Another piece that you may notice is "off/incorrect" in the PR is that we will guard the auto-config not on ReactivePulsarTemplate, but rather the Reactive client class. Because we are leaving `spring-pulsar` to contain the imperative and the reactive code (for now) we want  to still be able to back off when the reactive client is not on the classpath. Later we may split it into `spring-pulsar-reactive` but for now they will remain in the same module.

Also, if you are interested in the backing Reactive client, it is [here](https://github.com/apache/pulsar-client-reactive).

onobc:
Closing in favor of https://github.com/spring-projects-experimental/spring-pulsar/pull/175

PR #169, opened 2022-10-20
Add cache to DefaultReactivePulsarSenderFactory

cbornet:
Also refactored DefaultReactiveMessageSenderFactoryTests because it was mocking inside the reactive client and it's never a good idea to mock inside your dependencies.


onobc (approved):
LGTM

PR #170, opened 2022-10-20
Add reactive template operations to send multiple messages with a r.s.Publisher

cbornet:


spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -79,22 +91,44 @@ public void setSchema(Schema<T> schema) {
...
+	}
 
-		final ReactiveMessageSender<T> sender = createMessageSender(topic, message, messageRouter, customizer);
+	private Flux<MessageId> doSendMany(String topic, Publisher<T> messages,
+			MessageSpecBuilderCustomizer<T> messageSpecBuilderCustomizer, MessageRouter messageRouter,
+			ReactiveMessageSenderBuilderCustomizer<T> customizer) {
+		final String topicName = ReactiveMessageSenderUtils.resolveTopicName(topic, this.reactiveMessageSenderFactory);
+		this.logger.trace(() -> String.format("Sending reactive msg to '%s' topic", topicName));
+
+		if (this.schema != null) {
```
cbornet, line 103: If the template has a schema, we can create the message sender right away and use `ReactiveMessageSender ::sendMessages` to send them as a stream.
Otherwise we need to wait to get a message to create it and we can't share it between messages. So we create one each time and use `ReactiveMessageSender ::sendMessage` to send messages individually.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -79,22 +91,44 @@ public void setSchema(Schema<T> schema) {
...
+			ReactiveMessageSender<T> sender = this.reactiveMessageSenderFactory.createReactiveMessageSender(topic,
+					this.schema, messageRouter,
+					customizer == null ? Collections.emptyList() : Collections.singletonList(customizer));
+
+			Flux<MessageSpec<T>> map = Flux.from(messages)
```
onobc, line 108: **[TinyNit]** I think `messageSpecs` may be more informative than `map`.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
onobc: Thanks for the great explanation. I think this would be a great thing to add as a comment.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -79,22 +91,44 @@ public void setSchema(Schema<T> schema) {
...
-		this.logger.trace(() -> String.format("Sending reative msg to '%s' topic", topicName));
+		return doSendMany(topic, Mono.just(message), messageSpecBuilderCustomizer, messageRouter, customizer).single();
+	}
 
-		final ReactiveMessageSender<T> sender = createMessageSender(topic, message, messageRouter, customizer);
+	private Flux<MessageId> doSendMany(String topic, Publisher<T> messages,
+			MessageSpecBuilderCustomizer<T> messageSpecBuilderCustomizer, MessageRouter messageRouter,
+			ReactiveMessageSenderBuilderCustomizer<T> customizer) {
+		final String topicName = ReactiveMessageSenderUtils.resolveTopicName(topic, this.reactiveMessageSenderFactory);
+		this.logger.trace(() -> String.format("Sending reactive msg to '%s' topic", topicName));
```
onobc, line 101: **[TinyNit]** Here and in some of the other log statements we say 'msg' when it is more likely 'messages'.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -79,22 +91,44 @@ public void setSchema(Schema<T> schema) {
...
+
+		if (this.schema != null) {
+			ReactiveMessageSender<T> sender = this.reactiveMessageSenderFactory.createReactiveMessageSender(topic,
```
onobc, line 104: I would be in favor of just calling into `createMessageSender` as it hides a bit of ternary customizer below - it does the correct thing w/ the schema.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderOperations.java
```diff
@@ -45,6 +47,24 @@
 	 */
 	Mono<MessageId> send(String topic, T message);
 
+	/**
+	 * Sends multiple messages to the default topic in a reactive manner.
+	 * @param messages the messages to send
+	 * @return the ids assigned by the broker to the published messages in the same order
+	 * as they were sent
+	 */
+	Flux<MessageId> send(Publisher<T> messages);
```
onobc, line 56: What is the benefit of accepting publisher as input? 

I do understand that `Publisher` is the Reactivestreams API rather than Reactor (aka more generic). I wonder if mixing them may be a bit confusing? Or is this a common practice in other Reactive frameworks?

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/ReactivePulsarTemplateTests.java
```diff
@@ -69,10 +72,17 @@ void sendMessageWithSpecificSchemaTest() throws Exception {
...
-						.isEqualTo(foo);
+
+				List<Foo> foos = new ArrayList<>();
+				for (int i = 0; i < 10; i++) {
+					foos.add(new Foo("Foo-" + UUID.randomUUID(), "Bar-" + UUID.randomUUID()));
+				}
+				pulsarTemplate.send(Flux.fromIterable(foos)).subscribe();
+
+				for (int i = 0; i < 10; i++) {
+					assertThat(consumer.receiveAsync()).succeedsWithin(Duration.ofSeconds(3))
```
onobc, line 83: I do like that this also verifies the order contract. 

> @return the ids assigned by the broker to the published messages in the same order
	 * as they were sent

Does "sent" imply the order that I handed them to the API in my flux/publisher or the order they ended up being sent? I think this would almost always be true, but what provides this guarantee? What about failed message in the middle of the flux - and then it is retried maybe? Just playing devils' advocate.

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/ReactivePulsarTemplateTests.java
```diff
@@ -164,32 +177,36 @@ void sendMessageTest(String testName, SendTestArgs testArgs) throws Exception {
...
+						SendTestArgs.useSpecificTopic(true).useSenderCustomizer()),
 				arguments("sendReactiveMessageToSpecificTopicWithAllOptions", SendTestArgs.useSpecificTopic(true)
-						.useCustomRouter(true).useMessageCustomizer(true).useSenderCustomizer(true)));
+						.useCustomRouter().useMessageCustomizer().useSenderCustomizer()));
 	}
 
 	static final class SendTestArgs {
 
-		private boolean useSpecificTopic;
+		private final boolean useSpecificTopic;
```
onobc, line 209: ~Does this have to be final? The rule we use when to use final is when it is an immutable object w/ constructor injection.~ Ok, this is the immutable constructed portion of it - good eye.

onobc:
> We could also have an API to accept JDK9+ `Flow.Publisher`.
> 
> ```java
> 	Flux<MessageId> send(String topic, Flow.Publisher<T> messages);
> ```
> 
> It's just a wrapper to call `send(topic, JdkFlowAdapter.flowPublisherToFlux(messages)` so I wasn't sure if there's enough value. If you think it has, please tell me and I'll add it.

I would prefer starting out w/ the base one and then adding more variants/flavors/convenience as we go.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: Oups, IDEA auto-naming I forgot to change.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: Indeed. This is copy/pasted from imperative so they should be changed there as well.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderOperations.java
cbornet: Yes that's common practice.
See for instance in [R2DBC](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/reactive/ReactiveCrudRepository.html#saveAll-org.reactivestreams.Publisher-) or [Spring-Kafka](https://docs.spring.io/spring-kafka/api/org/springframework/kafka/core/reactive/ReactiveKafkaProducerTemplate.html#send(org.reactivestreams.Publisher)).
This allows the user to more easily mix between libs that may not use the same reactive streams impl.
A good application of the open/close principle / Postel law 😃

cbornet: We're also doing it in the Pulsar reactive client : https://github.com/apache/pulsar-client-reactive/pull/12

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: done

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/ReactivePulsarTemplateTests.java
cbornet: The reactive driver uses `flatMapSequential` on the `sendAsync` results. That will keep the order and should be tested in the reactive driver (not done atm).
I changed the logic when the template doesn't have a schema to also use `flatMapSequential`. I'll see if I can add a test to enforce the verification of the order.

cbornet: Failure of a mono in the flux will be handled as a usual r.s exception.
Maybe @lhotari can provide more details.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarSenderTemplate.java
```diff
@@ -98,25 +98,30 @@ private Flux<MessageId> doSendMany(String topic, Publisher<T> messages,
...
+			 * it between messages. So we create one each time and use
+			 * ReactiveMessageSender::sendMessage to send messages individually.
+			 */
+			ReactiveMessageSender<T> sender = createMessageSender(topic, null, messageRouter, customizer);
+			return Flux.from(messages).map(message -> getMessageSpec(messageSpecBuilderCustomizer, message))
+					.as(sender::sendMessages)
+					.doOnError(ex -> this.logger.error(ex,
+							() -> String.format("Failed to send messages to '%s' topic", topicName)))
+					.doOnNext(
+							msgId -> this.logger.trace(() -> String.format("Sent messages to '%s' topic", topicName)));
```
onobc, line 117: Hehe. This particular msg was correct before - as it was a single msg in `onNext`... We can change it later though.

onobc (approved):
Looks good! Thank you for another awesome contribution @cbornet .

PR #175, opened 2022-10-24
Add autoconfiguration for reactive Pulsar sender template

cbornet:


spring-pulsar-dependencies/build.gradle
```diff
@@ -28,5 +28,6 @@ dependencies {
 		api "com.google.protobuf:protobuf-java:$protobufJavaVersion"
 		api "org.apache.pulsar:pulsar-client-all:$pulsarVersion"
 		api "org.apache.pulsar:pulsar-client-reactive-adapter:$pulsarClientReactiveVersion"
+		api "org.apache.pulsar:pulsar-client-reactive-producer-cache-caffeine:$pulsarClientReactiveVersion"
```
onobc, line 31: This should be fine, but just sharing info...

The Reactive adapter brings in caffeine `2.9.3` but we bring in `3.1.1` currently. Gradle version alignment will kick in and pick the latest version - as long as the Reactive adapter is ok w/ that all will be fine. Just a heads up.

spring-pulsar-spring-boot-autoconfigure/build.gradle
```diff
@@ -9,7 +9,9 @@ dependencies {
 	annotationProcessor 'org.springframework.boot:spring-boot-autoconfigure-processor'
 	annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
-	api project (':spring-pulsar')
+	optional project (':spring-pulsar')
+	optional 'org.apache.pulsar:pulsar-client-reactive-producer-cache-caffeine'
```
onobc, line 13: Although the cache module brings in the adapter, I would be in favor of explicitly bringing in the adapter as optional here as well.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * Configuration properties for Spring for the Apache Pulsar reactive client.
+ * <p>
+ * Users should refer to Pulsar reactive client documentation for complete descriptions of
+ * these properties.
+ *
+ * @author Christophe Bornet
+ */
+@ConfigurationProperties(prefix = "spring.pulsar.reactive")
+public class PulsarReactiveProperties {
```
onobc, line 41: I think eventually we will want to harmonize some of the config props (Reactive / Imperative) as there is quite a bit of overlap. For now, we can go in as completely separate for simplicity.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
```diff
@@ -0,0 +1,320 @@
...
+
+		/**
+		 * Time before a message has to be acknowledged by the broker in milliseconds.
+		 */
+		private Duration sendTimeout = Duration.ofSeconds(30);
+
+		/**
+		 * Maximum number of pending messages for the producer.
+		 */
+		private int maxPendingMessages = 1000;
```
onobc, line 73: Typically, it is better to use big primitives (Integer vs. int) for config props values if you want to allow null (aka no setting). That being said, the existing PP do not follow that so we can revisit this later for both flavors of the props.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
```diff
@@ -0,0 +1,238 @@
...
+import org.springframework.pulsar.core.reactive.ReactivePulsarSenderFactory;
+import org.springframework.pulsar.core.reactive.ReactivePulsarSenderTemplate;
+
+/**
+ * Autoconfiguration tests for {@link PulsarReactiveAutoConfiguration}.
+ *
+ * @author Christophe Bornet
+ */
+@SuppressWarnings("unchecked")
+class PulsarReactiveAutoConfigurationTests {
```
onobc, line 55: One mental "TODO" I have is to add some PulsarProducerProperties as we have some gaps in coverage there currently. I think we have the same "TODO" here as well. The ServerPropertiesTests in SB are a good example of using the "bind" directly to test only the properties aspect of things (not the auto-config etc)

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
```diff
@@ -0,0 +1,238 @@
...
+		ReactivePulsarClient client = AdaptedReactivePulsarClientFactory.create((PulsarClient) null);
+		this.contextRunner.withBean("customReactivePulsarClient", ReactivePulsarClient.class, () -> client).run(
+				(context) -> assertThat(context).hasNotFailed().getBean(ReactivePulsarClient.class).isSameAs(client));
+	}
+
+	@Test
+	void customProducerCacheProviderIsRespected() throws Exception {
+		try (ProducerCacheProvider provider = new ConcurrentHashMapProducerCacheProvider()) {
+			this.contextRunner
+					.withBean("customReactivePulsarSenderTemplate", ProducerCacheProvider.class, () -> provider)
```
onobc, line 95: This (and a few tests below) name the bean "customReactivePulsarSenderTemplate" but its not that. :)

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
```diff
@@ -0,0 +1,238 @@
...
+					.run((context) -> assertThat(context).hasNotFailed().getBean(ProducerCacheProvider.class)
+							.isSameAs(provider));
```
onobc, line 97: It would probably be a good idea (and easy) to add a verification that the PCP provided cache is the one that is ultimately used.

spring-pulsar-spring-boot-autoconfigure/src/main/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
onobc, line 1: This should be under `test/resources...`

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for Apache Pulsar.
+ *
+ * @author Chris Bono
+ * @author Christophe Bornet
+ */
+@AutoConfiguration(after = PulsarAutoConfiguration.class)
+@ConditionalOnClass(ReactivePulsarSenderTemplate.class)
```
onobc, line 47: We actually want to make it conditional on the reactive client. If/when we split out the core spring-pulsar module we could base the condition on the RPT. Currently it will always be available since it is in `spring-pulsar`. 

I believe this is also why the `compileOnly` thing had to be done. I changed this locally and was able to put the deps. back as `api`.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+public class PulsarReactiveAutoConfiguration {
+
+	private final PulsarReactiveProperties properties;
+
+	public PulsarReactiveAutoConfiguration(PulsarReactiveProperties properties) {
+		this.properties = properties;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(ReactivePulsarClient.class)
```
onobc, line 58: **[NIT]** The default of COMB is the return type so we can simplify by removing the classname in the value attribute.

spring-pulsar/build.gradle
```diff
@@ -19,11 +18,13 @@ dependencies {
 	implementation 'com.fasterxml.jackson.core:jackson-core'
 	implementation 'com.fasterxml.jackson.core:jackson-databind'
 	implementation 'com.google.code.findbugs:jsr305'
+	compileOnly 'org.apache.pulsar:pulsar-client-reactive-adapter'
```
onobc, line 21: We can go back to `api` when we change the COMB in auto-config to the reactive client rather than reactive template class.

onobc (changes requested):
This is looking good @cbornet . A few comments to address. Thank you.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
cbornet: We already test below that DefaultReactivePulsarSenderFactory [uses the provided cache](https://github.com/spring-projects-experimental/spring-pulsar/pull/175/files/32cac1e3ee5de637a297dbc494da6a0f12551921#diff-893ec012124f1a540641a8ddbaff726756784d8c2b04d4e2d895e4cf777846b2R150).
Or by ultimately, do you mean it's the one used when calling `ReactivePulsarSenderFactory::createReactiveMessageSender`. That's unit tested in `DefaultReactiveMessageSenderFactoryTests` but not in integration.

spring-pulsar/build.gradle
cbornet: The idea was to not have spring-pulsar pull the reactive client. But that was maybe a bad idea. Setting it to `api.`

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
cbornet: Shouldn't it be both ReactivePulsarSenderTemplate and ReactivePulsarClient ?
If we have the reactive client but not spring-pulsar on the CP, this will fail...

cbornet: And with the comment https://github.com/spring-projects-experimental/spring-pulsar/pull/175#discussion_r1005889990, spring-pulsar will pull the reactive client so it's only spring-pulsar's presence we should test

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
cbornet: Yes. This could also be done PulsarAutoConfiguration 😃

spring-pulsar-spring-boot-autoconfigure/build.gradle
cbornet: With comment https://github.com/spring-projects-experimental/spring-pulsar/pull/175/files#r1005889990 it will also be pulled by spring-pulsar.
If we do it for the reactive adapter, shouldn't we do it also for the regular client ?

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
cbornet: done

spring-pulsar-spring-boot-autoconfigure/src/main/resources/mockito-extensions/org.mockito.plugins.MockMaker
cbornet: done

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
cbornet: done

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
cbornet: done

spring-pulsar-dependencies/build.gradle
onobc: Unrelated to this PR at this point, but should we shade the caffeine jar for the reactive-client (or is it `implementation` and not on the consumers classpath?)

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
onobc: Right, but we have some [logic in the auto-config](https://github.com/spring-projects-experimental/spring-pulsar/pull/175/files/32cac1e3ee5de637a297dbc494da6a0f12551921#diff-e405d49a17c8d841b416a53a1ffb8848feff7ec3f1feacec13a322811d4cfe55R81) that invokes the provider  that is not tested.

spring-pulsar/build.gradle
onobc: @cbornet sorry, I meant I was able to put it to `optional` w/ this change. Not `api`.

spring-pulsar-dependencies/build.gradle
cbornet: Caffeine jar is `api` in the reactive-client.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfigurationTests.java
cbornet: We test both cases where the [provider bean is absent](https://github.com/spring-projects-experimental/spring-pulsar/pull/175/files/32cac1e3ee5de637a297dbc494da6a0f12551921#diff-893ec012124f1a540641a8ddbaff726756784d8c2b04d4e2d895e4cf777846b2R214-R215) and where [it is present](https://github.com/spring-projects-experimental/spring-pulsar/pull/175/files/32cac1e3ee5de637a297dbc494da6a0f12551921#diff-893ec012124f1a540641a8ddbaff726756784d8c2b04d4e2d895e4cf777846b2R231-R232).
What is missing ?

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
onobc: > Shouldn't it be both ReactivePulsarSenderTemplate and ReactivePulsarClient ?
Yes, this would be safer to do. I can add this in polish commit during merge though.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveAutoConfiguration.java
onobc: > Yes. This could also be done PulsarAutoConfiguration 

😸 Agreed. I will update it there as well.

spring-pulsar-dependencies/build.gradle
onobc: We might consider shading it at some point.

onobc (approved):

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/e8b4ac1633e5d8b5ce33912607ea4a89d8394fd4

PR #178, opened 2022-10-27
Add ReactivePulsarConsumerFactory

cbornet:


spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarConsumerFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Pulsar reactive consumer factory interface.
+ *
+ * @param <T> payload type for the consumer.
+ * @author Christophe Bornet
+ */
+public interface ReactivePulsarConsumerFactory<T> {
+
+	ReactiveMessageConsumer<T> createConsumer(Schema<T> schema, ReactiveMessageConsumerSpec spec);
```
cbornet, line 31: Should we also add `ReactiveMessageConsumer<T> createConsumer(Schema<T> schema)` ?

onobc: I noticed this differs from the RPSF in that the producer variant does not allow user to pass in a spec but rather allows them to get the spec.

cbornet: Yes we can do it also for RSPF. RSPF is consistent with PulsarProducerFactory (ie PulsarProducerFactory doesn't have properties in its methods whereas PulsarConsumerFactory). I think we should add `properties` in the  PulsarProducerFactory methods.

onobc: Ahh ok. I see it now. 

>  I think we should add properties in the PulsarProducerFactory methods.

In the PPF there is already somewhat of a method signature explosion - adding `properties` would add to that. 

I would love for produce/consume be consistent in this area - one way or the other. Because of 1) the signature explosion and 2) changing the PPF would be quite a ripple I would prefer if we make the consumers follow the pattern of the producers. Then the global pattern would be:
- `<init>` gets the default spec/props
- expose the default spec/props via `get<Producer|Consumer><Spec|Props>`

Wdyt?

cbornet: So you can't customize props per Consumer ?

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageConsumerFactoryTests.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests for {@link DefaultReactivePulsarConsumerFactory}
+ *
+ * @author Christophe Bornet
+ */
+class DefaultReactiveMessageConsumerFactoryTests {
+
+	protected final Schema<String> schema = Schema.STRING;
```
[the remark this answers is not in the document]
cbornet: Done. I made it a constant actually.

spring-pulsar/src/main/java/org/springframework/pulsar/core/reactive/ReactivePulsarConsumerFactory.java
onobc: In the PPF the hook to customize is via the ProducerBuilderCustomizer. Should we do the same here?

cbornet: Oh yes. In that case, we do have a `ReactiveMessageSenderBuilderCustomizer`.

cbornet: So I'll add the `getReactiveMessageSenderSpec ` and `createConsumer(Schema<T> schema)`.
We should probably also have `createConsumer(Schema<T> schema)` for the imperative `PulsarConsumerFactory`.

cbornet: Done.
Also added ReactiveMessageSenderBuilderCustomizer for consistency (I'll do a PR to align the imperative API)

cbornet: Using a customizer is much better. It allows to set `maxInFlight`, the `cache` or `maxConcurrentSenderSubscriptions`

spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageConsumerFactoryTests.java
```diff
@@ -0,0 +1,70 @@
...
+ * Tests for {@link DefaultReactivePulsarConsumerFactory}
+ *
+ * @author Christophe Bornet
+ */
+class DefaultReactiveMessageConsumerFactoryTests {
+
+	private static final Schema<String> SCHEMA = Schema.STRING;
+
+	@Test
+	void createConsumer() {
```
onobc, line 42: Not super concerning, but we are missing the case w/ the null spec to `<init>` to verify the wrapping of the null->mutable spec. I can add during merge, or we can add next time we are here in the code though.

onobc (approved):
@cbornet I like where you landed on the API on the RPCF - it is simple and clean. LGTM.

PR #185, opened 2022-10-28
Add tests to DefaultReactiveMessageConsumerFactoryTests (#185)

onobc:
DO NOT MERGE: For discussion only.

@cbornet I was going to add a polish commit to recently merged PR (see [comment](https://github.com/spring-projects-experimental/spring-pulsar/pull/178#discussion_r1008572835)). However, when doing so I figured I wanted to run both of these styles by you before pushing a commit on top of yours w/o your approval / review. 

I wanted to add coverage for the case where the RPCF was created w/ a null spec. It basically doubles the test methods such that we check both `createConsumer()` w/ customizer and w/o customizer on both factories (one created w/ spec and one created w/o spec). 

I did it 2 ways in this proposal - w/ and w/o the use of `@Nested`. Which do you prefer? Or, would a single test just verifying the `<init> w/ null spec` be sufficient?

### DIFF
--- a/spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageConsumerFactoryTests.java
+++ b/spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageConsumerFactoryTests.java
@@ -27,44 +27,79 @@ import org.apache.pulsar.reactive.client.api.MutableReactiveMessageConsumerSpec;
 import org.apache.pulsar.reactive.client.api.ReactiveMessageConsumer;
 import org.apache.pulsar.reactive.client.api.ReactiveMessageConsumerSpec;
 import org.assertj.core.api.InstanceOfAssertFactories;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


spring-pulsar/src/test/java/org/springframework/pulsar/core/reactive/DefaultReactiveMessageConsumerFactoryNestedTests.java
```diff
@@ -0,0 +1,102 @@
...
+ * Tests for {@link DefaultReactivePulsarConsumerFactory}
+ *
+ * @author Christophe Bornet
+ * @author Chris Bono
+ */
+class DefaultReactiveMessageConsumerFactoryNestedTests {
+
+	private static final Schema<String> SCHEMA = Schema.STRING;
+
+	@Nested
```
onobc, line 44: Nested does not really shine in this small 2 tests in each group case. When there are many tests in the group it can really help visually group the tests. I am a fan of Nested though and use it quite often.

cbornet: +1 to use Nested

cbornet:
LGTM with Nested.

PR #186, opened 2022-10-29
Split sample apps into multi-modules

onobc:
This simplifies things when having multiple boot apps in single module. Also we want to add a reactive sample app(s) but they will have different dependencies - this allows that.


onobc:
@sobychacko I am going to push this one in as I want to get it into the reactive-preview branch so I can add a reactive example. Its just re-organizing the existing samples.

PR #187, opened 2022-10-30
Split sample apps and add reactive sample app

onobc:
This cherry picks the sample app split that was done in `main` branch. It then adds a sample reactive app in the 2nd commit. 

**NOTE:** The reactive app is just a starting place. @cbornet feel free to adjust however you see fit as we go through building out the Reactive pieces. 

**NOTE:** The name of the imperative apps are still `app1/2` and we should update this - but left the as-is for the scope of this proposal.

cc: @sobychacko


spring-pulsar-sample-apps/sample-reactive/src/main/java/org.springframework.pulsar.example/ReactiveSpringPulsarBootApp.java
```diff
@@ -0,0 +1,122 @@
...
+		ReactivePulsarConsumerFactory<Foo> reactivePulsarConsumerFactory(ReactivePulsarClient reactivePulsarClient) {
+			MutableReactiveMessageConsumerSpec spec = new MutableReactiveMessageConsumerSpec();
+			spec.setTopicNames(Collections.singletonList("sample-reactive-topic1"));
+			spec.setSubscriptionName("sample-reactive-sub1");
+			spec.setConsumerName("sample-reactive-consumer1");
+			return new DefaultReactivePulsarConsumerFactory<>(reactivePulsarClient, spec);
+		}
+
+		@Bean
+		ApplicationRunner listenPojo(ReactivePulsarConsumerFactory<Foo> reactiveConsumerFactory) {
```
onobc, line 75: **NOTE** ApplicationRunner executes before ApplicationReadyEvent fires.

PR #188, opened 2022-11-01
Use type safe structures to configure the consumer

cbornet:
* Use `ConsumerBuilderCustomizer` to customize the consumer when creating it
* Remove `BatchReceivePolicy` from the `createConsumer` params as it can now be configured through `ConsumerBuilderCustomizer`
* Add method `createConsumer(Schema<T> schema, Collection<String> topics, Map<String, String> properties,  List<ConsumerBuilderCustomizer<T>> customizers)` to be able to replace the topics and the consumer properties (the builder methods can only add to the default).
* Set `ackTimeoutMillis` to 1000 instead of 1 in tests as the ack timeout must not be below 1 second (and now it's checked by the code 🙂)

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
@@ -46,17 +46,19 @@ import org.springframework.util.CollectionUtils;
  */
 public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T> {
 
-	private final Map<String, Object> consumerConfig = new HashMap<>();
+	private final Map<String, Object> consumerConfig;
 
 	private final List<Consumer<T>> consumers = new ArrayList<>();
 
 	private final PulsarClient pulsarClient;
 
+	public DefaultPulsarConsumerFactory(PulsarClient pulsarClient) {
+		this(pulsarClient, Collections.emptyMap());
+	}
+
 	public DefaultPulsarConsumerFactory(PulsarClient pulsarClient, Map<String, Object> consumerConfig) {
 		this.pulsarClient = pulsarClient;
-		if (!CollectionUtils.isEmpty(consumerConfig)) {
-			this.consumerConfig.putAll(consumerConfig);
-		}
+		this.consumerConfig = Collections.unmodifiableMap(consumerConfig);
 	}
 
 	@Override
@@ -73,8 +75,8 @@ public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T>
 	public Consumer<T> createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
 			@Nullable Map<String, String> properties, @Nullable List<ConsumerBuilderCustomizer<T>> customizers)
 			throws PulsarClientException {
-		final ConsumerBuilder<T> consumerBuilder = this.pulsarClient.newConsumer(schema);
-		final Map<String, Object> config = new HashMap<>(this.consumerConfig);
+		ConsumerBuilder<T> consumerBuilder = this.pulsarClient.newConsumer(schema);
+		Map<String, Object> config = new HashMap<>(this.consumerConfig);
 
 		if (topics != null) {
 			config.put("topicNames", new HashSet<>(topics));
@@ -97,13 +99,13 @@ public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T>
 		}
 
 		if (config.containsKey("negativeAckRedeliveryBackoff")) {
-			final RedeliveryBackoff negativeAckRedeliveryBackoff = (RedeliveryBackoff) config
+			RedeliveryBackoff negativeAckRedeliveryBackoff = (RedeliveryBackoff) config
 					.get("negativeAckRedeliveryBackoff");
 			consumerBuilder.negativeAckRedeliveryBackoff(negativeAckRedeliveryBackoff);
 		}
 
 		if (config.containsKey("ackTimeoutRedeliveryBackoff")) {
-			final RedeliveryBackoff ackTimeoutRedeliveryBackoff = (RedeliveryBackoff) config
+			RedeliveryBackoff ackTimeoutRedeliveryBackoff = (RedeliveryBackoff) config
 					.get("ackTimeoutRedeliveryBackoff");
 			consumerBuilder.ackTimeoutRedeliveryBackoff(ackTimeoutRedeliveryBackoff);
 		}
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
@@ -17,11 +17,14 @@
 package org.springframework.pulsar.core;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
-import org.apache.pulsar.client.api.BatchReceivePolicy;
 import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.ConsumerBuilder;
 import org.apache.pulsar.client.api.DeadLetterPolicy;
@@ -30,6 +33,7 @@ import org.apache.pulsar.client.api.PulsarClientException;
 import org.apache.pulsar.client.api.RedeliveryBackoff;
 import org.apache.pulsar.client.api.Schema;
 
+import org.springframework.lang.Nullable;
 import org.springframework.util.CollectionUtils;
 
 /**
@@ -38,6 +42,7 @@ import org.springframework.util.CollectionUtils;
  * @param <T> underlying payload type for the consumer.
  * @author Soby Chacko
  * @author Alexander Preuß
+ * @author Christophe Bornet
  */
 public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T> {
 
@@ -45,7 +50,7 @@ public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T>
 
 	private final List<Consumer<T>> consumers = new ArrayList<>();
 
-	private PulsarClient pulsarClient;
+	private final PulsarClient pulsarClient;
 
 	public DefaultPulsarConsumerFactory(PulsarClient pulsarClient, Map<String, Object> consumerConfig) {
 		this.pulsarClient = pulsarClient;
@@ -55,58 +60,58 @@ public class DefaultPulsarConsumerFactory<T> implements PulsarConsumerFactory<T>
 	}
 
 	@Override
-	public Consumer<T> createConsumer(Schema<T> schema, Map<String, Object> propertiesToOverride)
-			throws PulsarClientException {
-
-		final ConsumerBuilder<T> consumerBuilder = this.pulsarClient.newConsumer(schema);
-
-		final Map<String, Object> properties = new HashMap<>(this.consumerConfig);
-		properties.putAll(propertiesToOverride);
-
-		if (!CollectionUtils.isEmpty(properties)) {
-			consumerBuilder.loadConf(properties);
-		}
-		Consumer<T> consumer = consumerBuilder.subscribe();
-		this.consumers.add(consumer);
-		return consumer;
+	public Consumer<T> createConsumer(Schema<T> schema) throws PulsarClientException {
+		return createConsumer(schema, null, null, Collections.emptyList());
 	}
 
 	@Override
-	public Consumer<T> createConsumer(Schema<T> schema, BatchReceivePolicy batchReceivePolicy,
-			Map<String, Object> propertiesToOverride) throws PulsarClientException {
+	public Consumer<T> createConsumer(Schema<T> schema, Collection<String> topics) throws PulsarClientException {
+		return createConsumer(schema, topics, null, Collections.emptyList());
+	}
 
+	@Override
+	public Consumer<T> createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
+			@Nullable Map<String, String> properties, @Nullable List<ConsumerBuilderCustomizer<T>> customizers)
+			throws PulsarClientException {
 		final ConsumerBuilder<T> consumerBuilder = this.pulsarClient.newConsumer(schema);
-		final Map<String, Object> properties = new HashMap<>(this.consumerConfig);
-		properties.putAll(propertiesToOverride);
+		final Map<String, Object> config = new HashMap<>(this.consumerConfig);
+
+		if (topics != null) {
+			config.put("topicNames", new HashSet<>(topics));
+		}
+		if (properties != null) {
+			config.put("properties", new TreeMap<>(properties));
+		}
 
 		// Remove deadLetterPolicy from the properties here and save it to re-apply after
 		// calling `loadConf` (https://github.com/apache/pulsar/issues/11646)
 		DeadLetterPolicy deadLetterPolicy = null;
-		if (properties.containsKey("deadLetterPolicy")) {
-			deadLetterPolicy = (DeadLetterPolicy) properties.remove("deadLetterPolicy");
+		if (config.containsKey("deadLetterPolicy")) {
+			deadLetterPolicy = (DeadLetterPolicy) config.remove("deadLetterPolicy");
 		}
 
-		if (!CollectionUtils.isEmpty(properties)) {
-			consumerBuilder.loadConf(properties);
-		}
+		consumerBuilder.loadConf(config);
 
 		if (deadLetterPolicy != null) {
 			consumerBuilder.deadLetterPolicy(deadLetterPolicy);
 		}
 
-		if (properties.containsKey("negativeAckRedeliveryBackoff")) {
-			final RedeliveryBackoff negativeAckRedeliveryBackoff = (RedeliveryBackoff) properties
+		if (config.containsKey("negativeAckRedeliveryBackoff")) {
+			final RedeliveryBackoff negativeAckRedeliveryBackoff = (RedeliveryBackoff) config
 					.get("negativeAckRedeliveryBackoff");
 			consumerBuilder.negativeAckRedeliveryBackoff(negativeAckRedeliveryBackoff);
 		}
 
-		if (properties.containsKey("ackTimeoutRedeliveryBackoff")) {
-			final RedeliveryBackoff ackTimeoutRedeliveryBackoff = (RedeliveryBackoff) properties
+		if (config.containsKey("ackTimeoutRedeliveryBackoff")) {
+			final RedeliveryBackoff ackTimeoutRedeliveryBackoff = (RedeliveryBackoff) config
 					.get("ackTimeoutRedeliveryBackoff");
 			consumerBuilder.ackTimeoutRedeliveryBackoff(ackTimeoutRedeliveryBackoff);
 		}
 
-		consumerBuilder.batchReceivePolicy(batchReceivePolicy);
+		if (!CollectionUtils.isEmpty(customizers)) {
+			customizers.forEach(customizer -> customizer.customize(consumerBuilder));
+		}
+
 		Consumer<T> consumer = consumerBuilder.subscribe();
 		this.consumers.add(consumer);
 		return consumer;
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
@@ -16,26 +16,62 @@
 
 package org.springframework.pulsar.core;
 
+import java.util.Collection;
+import java.util.List;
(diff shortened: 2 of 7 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -38,82 +41,176 @@
...
 		this.consumers.add(consumer);
 		return consumer;
 	}
 
-	public Map<String, Object> getConsumerConfig() {
+	public ConsumerConfigurationData<T> getConsumerConfig() {
 		return this.consumerConfig;
 	}
 
+	private static <T> void loadConf(ConsumerBuilder<T> consumer, ConsumerConfigurationData<T> conf) {
```
cbornet, line 103: This would need to be fully tested

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -240,15 +253,171 @@ else if (messageListener != null) {
...
 			}
 		}
 
+		/*
+		 * Method to work around the issue that ConsumerBuilder::loadConf sets JsonIgnored
+		 * fields of ConsumerConfigurationData to null. To be removed once
+		 * https://github.com/apache/pulsar/issues/11646 is fixed.
+		 */
+		@SuppressWarnings("unchecked")
+		private static <T> void loadConf(ConsumerBuilder<T> builder, Map<String, Object> properties) {
```
cbornet, line 281: this would need to be fully tested

onobc: Bummer - this is unfortunate. Could we not instead remove+remember the problematic properties, then call the standard loadConf, and finally apply the problem fields manually after the fact. It seems like it would reduce this additional area of mapping.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -240,15 +253,172 @@ else if (messageListener != null) {
 						.maxNumMessages(containerProperties.getMaxNumMessages())
 						.maxNumBytes(containerProperties.getMaxNumBytes())
 						.timeout(containerProperties.getBatchTimeoutMillis(), TimeUnit.MILLISECONDS).build();
+
+				Set<String> topicNames = (Set<String>) propertiesToConsumer.remove("topicNames");
```
onobc, line 257: If we end up going w/ this approach I think we should add a comment about why we are pulling these 2 fields out of the map (it is not too hard to figure this out below as they are 1st class method params but it could be helpful to have a simple comment as such).

onobc (changes requested):
@cbornet thanks for this code proposal. I am all for using the type-safe configuration options but I am wondering if the benefits are outweighing the drawbacks at this point.

#### Concern 1 - added field mappings
The manual mappings (aka the heroics) we are having to do in the DPMLC and the DPCF may outweigh the benefits of this type-safe change. I do realize some of this is due to bugs in the config data impl and will eventually be addressed. 


#### Concern 2 - is config data public?
In digging around for the client and producer config data side of things, I noticed that all of these config datas are in the "impl" package and are not exposed by the interfaces. They are only exposed by the default implementations. I believe this is the reason there is only a `loadConf(Map)` and not a `loadConf(<type>ConfigurationData)`. I think before going forward we need to be sure that the config data is considered part of the public API. It does not seem that way to me after digging in a bit. 

I reviewed all of the changes except for the tests/**. I want to wait until we talk about the above concerns before I invest in reviewing the tests (but mostly because it is late now 😄 )

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
cbornet: The problem is that if you have for instance a `deadLetterPolicy` in the current builder, calling `loadConf` will erase it. And we don't have access to the inner values of the builder to remember them before.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -38,82 +41,173 @@
...
 	}
 
 	@Override
-	public Consumer<T> createConsumer(Schema<T> schema, BatchReceivePolicy batchReceivePolicy,
-			Map<String, Object> propertiesToOverride) throws PulsarClientException {
-
+	public Consumer<T> createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
+			@Nullable Map<String, String> properties, @Nullable List<ConsumerBuilderCustomizer<T>> customizers)
+			throws PulsarClientException {
 		final ConsumerBuilder<T> consumerBuilder = this.pulsarClient.newConsumer(schema);
```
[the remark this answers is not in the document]
cbornet: Sure. I'll do a pass on all the `final`s that can be removed

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -38,82 +41,173 @@
...
 	}
 
-	public Map<String, Object> getConsumerConfig() {
+	public ConsumerConfigurationData<T> getConsumerConfig() {
 		return this.consumerConfig;
 	}
 
+	private static <T> void loadConf(ConsumerBuilder<T> consumer, ConsumerConfigurationData<T> conf) {
+		if (!CollectionUtils.isEmpty(conf.getTopicNames())) {
+			consumer.topics(conf.getTopicNames().stream().toList());
```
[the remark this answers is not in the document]
cbornet: Yes I know... Can't do much about it here... I believe ultimately we don't care about order in the consumer subscription and unicity could be achieved internally so accepting a `Collection` or maybe even an `Iterable` would have been better.
I'll see if we can make it better in the reactive client.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -90,8 +91,8 @@ public Admin getAdministration() {
 		return this.admin;
 	}
 
-	public Map<String, Object> buildConsumerProperties() {
-		return new HashMap<>(this.consumer.buildProperties());
+	public ConsumerConfigurationData<?> buildConsumerProperties() {
+		return this.consumer.buildProperties();
 	}
 
 	public Map<String, Object> buildClientProperties() {
```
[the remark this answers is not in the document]
cbornet: Indeed !

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
```diff
@@ -16,26 +16,65 @@
...
+			throws PulsarClientException;
 
-	Consumer<T> createConsumer(Schema<T> schema, BatchReceivePolicy batchReceivePolicy,
-			Map<String, Object> propertiesToOverride) throws PulsarClientException;
+	/**
+	 * Create a consumer.
+	 * @param schema the schema of the messages to be sent
+	 * @param topics the topics the consumer will subscribe to or {@code null} to use the
+	 * default topics
+	 * @param properties the properties to attach to the consumer or {@code null} to use
```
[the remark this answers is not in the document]
cbornet: I used the word `attach` from the definition in `ConsumerConfigurationData`.
I'll find a better formulation.
```java
    @ApiModelProperty(
            name = "properties",
            value = "A name or value property of this consumer.\n"
                    + "\n"
                    + "`properties` is application defined metadata attached to a consumer.\n"
                    + "\n"
                    + "When getting a topic stats, associate this metadata with the consumer stats for easier "
                    + "identification."
    )
    private SortedMap<String, String> properties = new TreeMap<>();
```

cbornet:
You're totally right that `ConsumerConfigurationData` is part of the client impl so it probably shouldn't be used outside of the pulsar-client internal.
There used to be a `ConsumerConfiguration` in the API but it's deprecated.
I see 2 options:
1/ Create a conf structure in spring-pulsar. It could have fluent setters and optional values. Mapping code would still be needed.
2/ Use a `ConsumerBuilderCustomizer` in the `DefaultPulsarConsumerFactory`
 instead of `ConsumerConfigurationData`. And also give the possibility to set default `topicNames` and `properties` that can be overriden later in `createConsumer`.
 
 My preference goes to option 2/

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
onobc: Yep, I remember the original issue. The difference here is ...

>  And we don't have access to the inner values of the builder to remember them before.

Is the only problem field the DLP? If so I would be in favor of digging that value out of the builder via reflection than have to do the mapping w/ all of the fields.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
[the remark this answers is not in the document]
onobc: 👍🏻 I think "replace" or "override" would do the trick (depending on what does of course).

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -38,82 +41,173 @@
...
 		this.consumers.add(consumer);
 		return consumer;
 	}
 
-	public Map<String, Object> getConsumerConfig() {
+	public ConsumerConfigurationData<T> getConsumerConfig() {
 		return this.consumerConfig;
 	}
 
+	private static <T> void loadConf(ConsumerBuilder<T> consumer, ConsumerConfigurationData<T> conf) {
```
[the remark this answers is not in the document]
onobc: > You're totally right that `ConsumerConfigurationData` is part of the client impl so it probably shouldn't be used outside of the pulsar-client internal. There used to be a `ConsumerConfiguration` in the API but it's deprecated. I see 3 options: 1/ Create a conf structure in spring-pulsar. It could have fluent setters and optional values. Mapping code would still be needed. 2/ Use a `ConsumerBuilderCustomizer` in the `DefaultPulsarConsumerFactory` instead of `ConsumerConfigurationData`. And also give the possibility to set default `topicNames` and `properties` that can be overriden later in `createConsumer`. 3/ Create a `DefaultPulsarConsumerFactoryBuilder` that can take set every prop individually.
> 
> My preference goes to option 2/

@cbornet 😞 Bummer. Thx for checking into that though. 

Of the above options I would go for 2) as well. I think let's get consensus on the signatures for `DefaultPulsarConsumerFactory.<init>` and `PulsarConsumerFactory` before moving forward though. 

There is also another option, leave the construction of DPCF using a map. I believe the primary use case is that the auto-configured DPCF will be used. Also, the auto-configuration hides this from the end user (`PulsarProperties->Map->DPCF.<init>`. So the type-safe `<init>` is not as big of a win as the type-safe `createConsumer` method w/ use of the customizers.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
cbornet: No DLP is not the only one. I can try to do the save/load/replace trick for all of them

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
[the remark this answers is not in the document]
onobc: **[NIT]** This is still "attach". I think going back to the original name of `propertiesToOverride` will clarify this. 

Also, I just realized that we were doing a "merge/override" before rather than a wholesale "replace". I believe we still want to allow them to override specific props rather than replace the whole map. 

In that case, the `CB.properties` does the right things (via `Map.putAll`) and we should be able to pull the properties param out altogether, no?

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
```diff
@@ -16,26 +16,62 @@
...
+	 * the default properties
+	 * @param customizers the optional list of customizers to apply to the consumer
+	 * builder
+	 * @return the consumer
+	 * @throws PulsarClientException if any error occurs
+	 */
+	Consumer<T> createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
```
onobc, line 67: It may be helpful to enhance the comments here to include the fact (at a high level) that we set topics via the param and if you set them on the builder it will append to the default topics, not replace them.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -55,58 +60,58 @@ public DefaultPulsarConsumerFactory(PulsarClient pulsarClient, Map<String, Objec
...
-		this.consumers.add(consumer);
-		return consumer;
+	public Consumer<T> createConsumer(Schema<T> schema) throws PulsarClientException {
+		return createConsumer(schema, null, null, Collections.emptyList());
 	}
 
 	@Override
-	public Consumer<T> createConsumer(Schema<T> schema, BatchReceivePolicy batchReceivePolicy,
-			Map<String, Object> propertiesToOverride) throws PulsarClientException {
+	public Consumer<T> createConsumer(Schema<T> schema, Collection<String> topics) throws PulsarClientException {
```
onobc, line 68: @cbornet I have not thought about this before and do not know the answer but do the nullity annotations inherit from the overridden signature? I see you have them below but not here.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -55,58 +60,58 @@ public DefaultPulsarConsumerFactory(PulsarClient pulsarClient, Map<String, Objec
...
+		return createConsumer(schema, topics, null, Collections.emptyList());
+	}
 
+	@Override
+	public Consumer<T> createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
+			@Nullable Map<String, String> properties, @Nullable List<ConsumerBuilderCustomizer<T>> customizers)
+			throws PulsarClientException {
 		final ConsumerBuilder<T> consumerBuilder = this.pulsarClient.newConsumer(schema);
```
onobc, line 76: **[NIT]** Please remove final 😸 (here and below).

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -38,14 +42,15 @@
...
 
 	private final Map<String, Object> consumerConfig = new HashMap<>();
 
 	private final List<Consumer<T>> consumers = new ArrayList<>();
 
-	private PulsarClient pulsarClient;
+	private final PulsarClient pulsarClient;
 
 	public DefaultPulsarConsumerFactory(PulsarClient pulsarClient, Map<String, Object> consumerConfig) {
 		this.pulsarClient = pulsarClient;
```
onobc, line 56: Wdyt about making the default config map immutable? I think it would be a good idea. Also, we can advertise this fact in `PCF.getConsumerConfig`.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -240,15 +253,172 @@ else if (messageListener != null) {
...
 			}
 			catch (PulsarClientException e) {
 				DefaultPulsarMessageListenerContainer.this.logger.error(e, () -> "Pulsar client exceptions.");
 			}
 		}
 
+		/*
+		 * Method to work around the issue that ConsumerBuilder::loadConf sets JsonIgnored
+		 * fields of ConsumerConfigurationData to null. To be removed once
+		 * https://github.com/apache/pulsar/issues/11646 is fixed.
```
onobc, line 276: Is this actually the related issue? I know this is the cause of DeadLetterPolicy hackery - maybe they are the same underlying issue?

onobc (changes requested):

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
cbornet: Actually we can't call `loadConf` because there will be a crash if DLP was set before on the builder.
So we don't have a choice but to map manually. We can remove this when `loadConf` is fixed.
IIUC `BeanUtils.copyProperties` copies fields with the same name from one struct to the other which is not what is done here.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
cbornet: Yes there's also a crash when you have a DLP in the builder. I'll update the comment.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
cbornet: It seems to me that the signatures are the same regarding `@Nullable` ?

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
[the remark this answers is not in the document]
cbornet: No that was already a replace. Here, it's the consumer properties as in `ConfigurationConsumerData.properties`, not the properties (previously called `propertiesToOverride `) from which we do `loadConf`.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
```diff
@@ -16,26 +16,62 @@
...
+	 * @param schema the schema of the messages to be sent
+	 * @return the consumer
+	 * @throws PulsarClientException if any error occurs
+	 */
+	Consumer<T> createConsumer(Schema<T> schema) throws PulsarClientException;
+
+	/**
+	 * Create a consumer.
+	 * @param schema the schema of the messages to be sent
+	 * @param topics the topics the consumer will subscribe to builder
```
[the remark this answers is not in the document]
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarConsumerFactory.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
cbornet: done via `Collections.unmodifiable`.
Or do you consider another immutable impl ?

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
onobc: Thats what I would have used as well. 👍🏻

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
onobc: Ahh yes they are consistent (interface + impl). I think `topics` should be `@Nullable` as it is in the other methods. I will adjust in polish commit.

onobc:
Superseded by https://github.com/spring-projects-experimental/spring-pulsar/pull/192

Issue #119, opened 2022-09-16
Consider using Duration / DataSize where applicable


onobc:
We are somewhat inconsistent w/ some of our data size and duration based properties. In some cases we do not name the var such that the unit can be determined. In some cases we name it w/ unit suffix of `ms` and in some `millis`. 

Lets scan through all of our properties and make this consistent. 
- PulsarProperties
- PulsarContainerProperties
- ???

_Originally posted by @onobc in https://github.com/spring-projects-experimental/spring-pulsar/pull/118#discussion_r972580211_

cbornet:
I will use Duration for the reactive API.

onobc:
Lets call this one closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/b4203bedc8dc2f7bd2783231f12a906a0bf5710b

Issue #176, opened 2022-10-24
Add config props unit tests


onobc:
We currently have primary coverage of both PulsarProperties and PulsarReactiveProperties. However, it would be a good ideal to create a 1st class unit test for each that focuses on the properties logic only. 

A good example is the [ServerPropertiesTests](https://github.com/spring-projects/spring-boot/blob/f11945b9618a97d2a8ec2d2d7b3ac893997e3d1a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/ServerPropertiesTests.java#L582) in SB where they do a direct bind on a map of properties to a config prop class.

This also opens the door to adding defaults tests to make sure that the default of the backing clients are in line w/ the defaults in the config props. See [CassandraPropertiesTests](https://github.com/spring-projects/spring-boot/blob/f11945b9618a97d2a8ec2d2d7b3ac893997e3d1a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cassandra/CassandraPropertiesTests.java#L42) for example. 

cc: @cbornet

onobc:
Let's call this one closed @cbornet . Any ones we missed we can add under a separate issue etc.. Thanks for knocking this out!

PR #191, opened 2022-11-03
Use Duration, DataSize and boxed types in PulsarProperties

cbornet:
See #119

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarAnnotationDrivenConfiguration.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarAnnotationDrivenConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.pulsar.autoconfigure;
 
+import java.time.Duration;
+
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -28,6 +30,7 @@ import org.springframework.pulsar.config.PulsarListenerBeanNames;
 import org.springframework.pulsar.core.PulsarConsumerFactory;
 import org.springframework.pulsar.listener.PulsarContainerProperties;
 import org.springframework.pulsar.observation.PulsarListenerObservationConvention;
+import org.springframework.util.unit.DataSize;
 
 import io.micrometer.observation.ObservationRegistry;
 
@@ -62,8 +65,9 @@ public class PulsarAnnotationDrivenConfiguration {
 		PulsarProperties.Listener listenerProperties = this.pulsarProperties.getListener();
 		map.from(listenerProperties::getSchemaType).to(containerProperties::setSchemaType);
 		map.from(listenerProperties::getAckMode).to(containerProperties::setAckMode);
-		map.from(listenerProperties::getBatchTimeoutMillis).to(containerProperties::setBatchTimeoutMillis);
-		map.from(listenerProperties::getMaxNumBytes).to(containerProperties::setMaxNumBytes);
+		map.from(listenerProperties::getBatchTimeout).asInt(Duration::toMillis)
+				.to(containerProperties::setBatchTimeoutMillis);
+		map.from(listenerProperties::getMaxNumBytes).asInt(DataSize::toBytes).to(containerProperties::setMaxNumBytes);
 		map.from(listenerProperties::getMaxNumMessages).to(containerProperties::setMaxNumMessages);
 
 		return new ConcurrentPulsarListenerContainerFactory<>(consumerFactoryProvider.getIfAvailable(),
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
@@ -22,7 +22,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
-import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 import org.apache.pulsar.client.api.CompressionType;
@@ -41,6 +40,7 @@ import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.pulsar.listener.AckMode;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
+import org.springframework.util.unit.DataSize;
 
 /**
  * Configuration properties for Spring for Apache Pulsar.
@@ -50,6 +50,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Soby Chacko
  * @author Alexander Preuß
+ * @author Christophe Bornet
  */
 @ConfigurationProperties(prefix = "spring.pulsar")
 public class PulsarProperties {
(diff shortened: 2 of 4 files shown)


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -111,12 +112,12 @@ public static class Consumer {
...
 		 * Comma-separated list of topics the consumer subscribes to.
 		 */
-		private String[] topics;
+		private Set<String> topics;
 
 		/**
 		 * Pattern for topics the consumer subscribes to.
 		 */
-		private String topicsPattern;
+		private Pattern topicsPattern;
```
cbornet, line 120: Used Pattern here since SB knows how to bind it

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -111,12 +112,12 @@ public static class Consumer {
 		/**
 		 * Comma-separated list of topics the consumer subscribes to.
 		 */
-		private String[] topics;
+		private Set<String> topics;
```
cbornet, line 115: Used `Set` here since SB knows how to bind it and `ConsumerBuilder` takes a `Set`

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
```diff
@@ -210,7 +210,7 @@ void customProducerInterceptorsOrderedProperly() {
 	void consumerBatchPropertiesAreHonored() {
 		contextRunner
 				.withPropertyValues("spring.pulsar.listener.max-num-messages=10",
-						"spring.pulsar.listener.max-num-bytes=101", "spring.pulsar.listener.batch-timeout-millis=50")
+						"spring.pulsar.listener.max-num-bytes=101B", "spring.pulsar.listener.batch-timeout=50ms")
```
sobychacko, line 213: Should we document these new formats in the doc?

sobychacko:
@cbornet Overall, I like the approach that you took. Thank you for doing this cleanup. I added a minor doc comment above.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarAutoConfigurationTests.java
onobc: These should be self documenting in that they are config props and the doc is auto-generated.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarAnnotationDrivenConfiguration.java
```diff
@@ -62,8 +65,9 @@ ConcurrentPulsarListenerContainerFactory<?> pulsarListenerContainerFactory(
 		PulsarProperties.Listener listenerProperties = this.pulsarProperties.getListener();
 		map.from(listenerProperties::getSchemaType).to(containerProperties::setSchemaType);
 		map.from(listenerProperties::getAckMode).to(containerProperties::setAckMode);
-		map.from(listenerProperties::getBatchTimeoutMillis).to(containerProperties::setBatchTimeoutMillis);
-		map.from(listenerProperties::getMaxNumBytes).to(containerProperties::setMaxNumBytes);
+		map.from(listenerProperties::getBatchTimeout).asInt(Duration::toMillis)
```
onobc, line 68: @cbornet good pragmatic call on stopping the conversion here (ie leave the downstream things still using the previous types) as that would have introduced quite the ripple. We can adjust those usages as we re-encounter those code areas (eg the ListenerContainerProperties in this case).

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -131,25 +132,23 @@ public static class Consumer {
...
 		 */
-		private long acknowledgementsGroupTimeMicros = TimeUnit.MILLISECONDS.toMicros(100);
+		private Duration acknowledgementsGroupTime = Duration.ofMillis(100);
 
 		/**
-		 * Delay before re-delivering messages that have failed to be processed in
-		 * microseconds.
+		 * Delay before re-delivering messages that have failed to be processed.
 		 */
-		private long negativeAckRedeliveryDelayMicros = TimeUnit.MINUTES.toMicros(1);
```
onobc, line 146: There is a reference to this property in `spring-pulsar-docs/src/main/asciidoc/pulsar.adoc` (lines 941,951). Can you take a scan through that for any property name changes in here? We have too many hardcoded code samples in the doc that we intend to migrate out to source code snippets at some point - in the meantime we just have to search manually when we rename things.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -375,63 +374,65 @@ public void setRegexSubscriptionMode(RegexSubscriptionMode regexSubscriptionMode
...
-			map.from(this::getTopics).as(Set::of).to(properties.in("topicNames"));
-			map.from(this::getTopicsPattern).as(Pattern::compile).to(properties.in("topicsPattern"));
+			map.from(this::getTopics).to(properties.in("topicNames"));
+			map.from(this::getTopicsPattern).to(properties.in("topicsPattern"));
 			map.from(this::getSubscriptionName).to(properties.in("subscriptionName"));
 			map.from(this::getSubscriptionType).to(properties.in("subscriptionType"));
 			map.from(this::getReceiverQueueSize).to(properties.in("receiverQueueSize"));
-			map.from(this::getAcknowledgementsGroupTimeMicros).to(properties.in("acknowledgementsGroupTimeMicros"));
-			map.from(this::getNegativeAckRedeliveryDelayMicros).to(properties.in("negativeAckRedeliveryDelayMicros"));
+			map.from(this::getAcknowledgementsGroupTime).as(it -> it.toNanos() / 1000)
```
onobc, line 427: This (the PR as a whole) is a change that would benefit from #176 being in place beforehand. Not saying to go do that for this PR - just saying out loud. 

I don't think these changes are too risky though and we also have fair test coverage (indirectly). I do remember some issue w/ the mapper not doing what I expected it to do when using `as` (even though we said "only apply when non-null" it was still executing). May be worth checking this particular case for a property using `as`.

onobc (changes requested):
I'm so glad you did this @cbornet . Thank you. Minor things to address.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
cbornet: Yes I think adding tests is the next step.

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
cbornet: done

PR #192, opened 2022-11-04
[Collaborated] Use type safe structures to configure the consumer 

onobc:
This PR is https://github.com/spring-projects-experimental/spring-pulsar/pull/188 with an additional collaboration commit to reduce the amount of mapping required in DPMLC. 

cc: @cbornet @sobychacko

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/ConsumerBuilderConfigurationUtil.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/ConsumerBuilderConfigurationUtil.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.core;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import org.apache.pulsar.client.api.ConsumerBuilder;
+import org.apache.pulsar.client.api.DeadLetterPolicy;
+import org.apache.pulsar.client.api.RedeliveryBackoff;
+import org.apache.pulsar.client.impl.ConsumerBuilderImpl;
+import org.apache.pulsar.client.impl.conf.ConsumerConfigurationData;
+
+/**
+ * Utility methods to help load configuration into a {@link ConsumerBuilder}.
+ * <p>
+ * The main purpose is to work around the underlying
+ * <a href="https://github.com/apache/pulsar/issues/11646">Pulsar issue</a> where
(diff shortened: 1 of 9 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -282,145 +275,27 @@ else if (messageListener != null) {
...
+			DeadLetterPolicy deadLetterPolicyDefault = confData.getDeadLetterPolicy();
+			confData.setDeadLetterPolicy(null);
+
+			// Now that DLP has been stripped from properties it is safe to call
+			// builder.loadConf
+			builder.loadConf(properties);
+
+			// Prefer the DLP from the container properties over any default DLP set on
+			// the consumer factory.
+			if (deadLetterPolicy != null) {
```
cbornet, line 294: I think all `JsonIgnored` properties of `ConsumerConfigurationData` should also be reapplied bc they are reset to `null` during `loadConf`.
Maybe we could have a common utility method for DPCF and DPLMC.

onobc: Ha! Yeh I dug into this last night and was thinking of doing both (util and extend the props). I had a comment queued up that looked like:

> A key takeaway that I just realized is that whatever properties we support in @PulsarListener via 
> https://github.com/spring-projects-experimental/spring-pulsar/blob/b4203bedc8dc2f7bd2783231f12a906a0bf5710b/spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListenerAnnotationBeanPostProcessor.java#L365-L367
> that also happen to be one of the @JsonIgnore properties in CCD must also be handled here specially. 
> 
> That list is pretty much as follows so we are clear for now w/ the current fields we are handling. 
> 
> This caveat was already present prior to this PR though. Just saying out loud. 
> 
> @JsonIgnore
> private MessageListener<T> messageListener;
> 
>   @JsonIgnore
>   private ConsumerEventListener consumerEventListener;
> 
>   @JsonIgnore
>   private RedeliveryBackoff negativeAckRedeliveryBackoff;
> 
>   @JsonIgnore
>   private RedeliveryBackoff ackTimeoutRedeliveryBackoff;
> 
>   @JsonIgnore
>   private transient MessagePayloadProcessor payloadProcessor;
> 
>   @JsonIgnore
>   private BatchReceivePolicy batchReceivePolicy;
> 
> @JsonIgnore
> private DeadLetterPolicy deadLetterPolicy;

So if we were to do a common utility we would want to cover all these fields. However, because the ones that we are currently supporting in the PulsarListener are being handled I wanted to push that off as we have enough going on in this PR already. Of course, if one were to use this w/o the PL and pass in one of the unhandled properties it would not be set. 

In summary - I would recommend getting this PR merged and then iterate on it from there. Wdyt?

cbornet: OK. LGTM 👍

spring-pulsar/src/main/java/org/springframework/pulsar/core/ConsumerBuilderConfigurationUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+	private ConsumerBuilderConfigurationUtil() {
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <T> void loadConf(ConsumerBuilder<T> builder, Map<String, Object> properties) {
+
+		ConsumerConfigurationData<T> builderConf = ((ConsumerBuilderImpl<T>) builder).getConf();
+
+		// Remove and remember problem fields from input props and builder
```
onobc, line 49: We could easily abstract out this code. I left it as-is for now.

PR #195, opened 2022-11-06
PulsarAdministration now respects auth properties

### LINKED ISSUE #194: Unable to configure authentication on admin client
### Problem
An app using `spring-pulsar` that defines `PulsarTopic` beans will fail to start when the admin required authentication. 

### Cause
The `PulsarAdmin` (admin client) does not convert the `authPluginClassName` and `authParams` into an authentication object like the `PulsarClient` does.

onobc:
This code proposal provides a workaround for the fact that the PulsarAdminBuilder does not respect the auth properties specified in the config map during a `loadConf`. 

### Details
* Commit 1: add test to show the breakage
* Commit 2: fix the breakage
* Commit 3: fix unrelated issue in properties assertion

### Next steps
* Submit fix to Apache Pulsar

Fixes #194

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
@@ -51,6 +51,7 @@ import org.springframework.util.unit.DataSize;
  * @author Soby Chacko
  * @author Alexander Preuß
  * @author Christophe Bornet
+ * @author Chris Bono
  */
 @ConfigurationProperties(prefix = "spring.pulsar")
 public class PulsarProperties {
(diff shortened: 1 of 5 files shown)


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1609,9 +1610,9 @@ public void setTlsProtocols(Set<String> tlsProtocols) {
 		}
 
 		public Map<String, Object> buildProperties() {
-			if (!StringUtils.hasText(this.getAuthParams()) && !CollectionUtils.isEmpty(this.getAuthentication())) {
+			if (StringUtils.hasText(this.getAuthParams()) && !CollectionUtils.isEmpty(this.getAuthentication())) {
```
onobc, line 1613: This was another bug that was unrelated to the auth props not being respected, but ultimately this prevented any auth props from being set at all.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarPropertiesTests.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.boot.context.properties.bind.Binder;
+import org.springframework.boot.context.properties.source.ConfigurationPropertySource;
+import org.springframework.boot.context.properties.source.MapConfigurationPropertySource;
+
+/**
+ * Unit tests for {@link PulsarProperties}.
+ *
+ * @author Chris Bono
+ */
+public class PulsarPropertiesTests {
```
onobc, line 39: **NOTE:** This is the beginning of the config props tests (#176)

sobychacko (approved):

sobychacko:
Looks good. Merged upstream.

Issue #117, opened 2022-09-15
Support AOT/native mode


onobc:
We should make sure that we can support running in AOT/native mode.

PR #202, opened 2022-11-14
Spring Pulsar AOT - initial changes

sobychacko:
Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/117

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/aot/PulsarRuntimeHints.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/aot/PulsarRuntimeHints.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.aot;
+
+import java.util.HashSet;
+import java.util.TreeMap;
+import java.util.stream.Stream;
+
+import org.apache.pulsar.client.admin.internal.OffloadProcessStatusImpl;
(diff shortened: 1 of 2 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/aot/PulsarRuntimeHints.java
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.aot;
+
+import java.util.HashSet;
+import java.util.TreeMap;
+import java.util.stream.Stream;
+
+import org.apache.pulsar.client.admin.internal.OffloadProcessStatusImpl;
```
onobc, line 23: I wonder if we should use the "string" version of the API for any Pulsar "internal" classes?

spring-pulsar/src/main/java/org/springframework/pulsar/aot/PulsarRuntimeHints.java
```diff
@@ -0,0 +1,92 @@
...
+ * {@link RuntimeHintsRegistrar} for Spring for Apache Pulsar.
+ *
+ * @author Soby Chacko
+ */
+public class PulsarRuntimeHints implements RuntimeHintsRegistrar {
+
+	@Override
+	public void registerHints(RuntimeHints hints, @Nullable ClassLoader classLoader) {
+		ReflectionHints reflectionHints = hints.reflection();
+		Stream.of(HashSet.class, TreeMap.class, Authentication.class, AuthenticationDataProvider.class,
```
onobc, line 54: Do. you mind adding a 1 liner comment about each section describing what/why etc? I think it may be priceless when we look back in here in like 6months and say "wuuuttt?" 😄

onobc (changes requested):
@sobychacko LGTM. Couple of minor comments to address. This is exciting. Nice work!

We should add a brief section in docs after this is merged.

spring-pulsar/src/main/java/org/springframework/pulsar/aot/PulsarRuntimeHints.java
sobychacko: We could do that, but what benefit it brings? If Pulsar client moves things around, we have to change in either case (string or type).

onobc: Yeh, I guess if they change things we will know quicker as the build will break on compile. this may actually be better than not knowing about it for N time period.

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/3e96769cefc285997a0753c10372516a885ec649

PR #203, opened 2022-11-15
Add support for ReactivePulsarListener annotation

cbornet:


spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpointRegistry.java
```diff
@@ -0,0 +1,265 @@
...
+			Assert.state(!this.listenerContainers.containsKey(id),
+					"Another endpoint is already registered with id '" + subscriptionName + "'");
+			C container = createListenerContainer(endpoint, factory);
+			this.listenerContainers.put(id, container);
+		}
+	}
+
+	protected C createListenerContainer(E endpoint, GenericPulsarListenerContainerFactory<? extends C, E> factory) {
+
+		if (endpoint instanceof MethodPulsarListenerEndpoint<?> mkle) {
```
cbornet, line 137: Oups. Something to do here

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpoint.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @author Christophe Bornet
+ */
+public interface GenericPulsarListenerEndpoint {
+
+	@Nullable
+	String getId();
+
+	@Nullable
+	String getSubscriptionName();
```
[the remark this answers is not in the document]
sobychacko: Also in the `ConcurrentPulsarListenerContainerFactory`.

sobychacko:
Wow! This looks very thorough. It is a large PR, though. I glanced through the code, looks good, will do a detailed review later. I see the places where you try to combine with the imperative model. I will let @onobc take a look as well. Thanks for this great work.

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,785 @@
...
+				};
+			}
+		}
+	}
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		this.registrar.setBeanFactory(this.beanFactory);
+
+		this.beanFactory.getBeanProvider(PulsarListenerConfigurer.class)
```
cbornet, line 207: It gets the same bean as the imperative. Is it OK ?

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,785 @@
...
+ * @author Christophe Bornet
+ * @see ReactivePulsarListener
+ * @see EnablePulsar
+ * @see PulsarListenerConfigurer
+ * @see PulsarListenerEndpointRegistrar
+ * @see ReactivePulsarListenerEndpointRegistry
+ * @see ReactivePulsarListenerEndpoint
+ * @see MethodReactivePulsarListenerEndpoint
+ */
+public class ReactivePulsarListenerAnnotationBeanPostProcessor<V>
```
cbornet, line 120: A lot of this should be mutualized with `PulsarListenerAnnotationBeanPostProcessor` I guess.

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpointRegistry.java
cbornet: Actually this seems to be dead code as the only call  of `EndpointHandlerMethod`'s constructor is inside this check. Probably some leftover from spring-kafka.
I propose to remove `EndpointHandlerMethod`. OK with that ?

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpoint.java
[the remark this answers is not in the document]
cbornet: I mean it could be put in `PulsarListenerEndpoint` and removed from `GenericPulsarListenerEndpoint`.
Currently we need it in `GenericPulsarListenerEndpoint` because we do:
```java
                String subscriptionName = endpoint.getSubscriptionName();
		String id = endpoint.getId();

		Assert.hasText(subscriptionName, "Endpoint id must not be empty");

		synchronized (this.listenerContainers) {
			Assert.state(!this.listenerContainers.containsKey(id),
					"Another endpoint is already registered with id '" + subscriptionName + "'");
			C container = createListenerContainer(endpoint, factory);
			this.listenerContainers.put(id, container);
		}
```
in `GenericPulsarListenerEndpointRegistry`.
But I don't understand why we check `subscriptionName` and then use `id` as container key. Shouldn't we only check `id` ? I looked at spring-kafka and they only check `id`.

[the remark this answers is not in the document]
sobychacko: We need it [here](https://github.com/spring-projects-experimental/spring-pulsar/blob/main/spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java#L70) for the concurrent factory to extract the subscription name from the `PulsarListener` annotation provided as listener properties.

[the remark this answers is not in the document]
sobychacko: Both `id` and `subsctiptionName` are not mandatory from a `PulsarListener` standpoint. If they are not present, the framework auto-generates them. Granted, the auto-generation of the subscription name only makes sense for an exclusive consumer; for anything else, the use might want to provide the sub name.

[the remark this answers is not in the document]
sobychacko: With that said, we could refactor this area if there is room for further improvement.

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListenerAnnotationBeanPostProcessor.java
sobychacko: I think it is OK to do so, as we are not expecting or suggesting mixing the two paradigms in the same app.

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListenerAnnotationBeanPostProcessor.java
sobychacko: Sounds like a good idea.

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpointRegistry.java
onobc: Yeh, this section of the code is doing quite a bit of things and when we find the dead code that was used in `spring-kafka` but not here in `spring-pulsar` too am in favor of pulling. 

That being said, I am also in favor of changing as little as "live" imperative code in this particular PR as its already big. I would much rather see a list of todos for a small subsequent PR. Not talking directly about this change, but just in general. I am about to review the rest of the PR now.

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListenerAnnotationBeanPostProcessor.java
onobc: Agreed. Once the "dust settles" from this PR lets come back and see what makes sense to harmonize.

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListener.java
```diff
@@ -0,0 +1,156 @@
...
+	 */
+	String subscriptionName() default "";
+
+	/**
+	 * Pulsar subscription type for this listener.
+	 * @return the {@code subscriptionType} for this listener
+	 */
+	SubscriptionType subscriptionType() default SubscriptionType.Exclusive;
+
+	SchemaType schemaType() default SchemaType.NONE;
```
onobc, line 79: **[NIT]** Do you mind adding javadoc here (as well as the imperative one that was already missing it)?

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListener.java
```diff
@@ -0,0 +1,156 @@
...
+	 * obtain the value.
+	 * <p>
+	 * SpEL {@code #{...}} and property place holders {@code ${...}} are supported.
+	 * @return true to auto start, false to not auto start.
+	 */
+	String autoStartup() default "";
+
+	/**
+	 * Activate stream consumption.
+	 * @return whether this listener is a streaming one.
```
onobc, line 117: In layman's terms, what is the result if this is `false`? Or put another way, what does a "streaming one" mean?

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpoint.java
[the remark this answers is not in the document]
onobc: I am all for refactoring - can we can do it in another PR though?

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/ReactivePulsarListenerEndpoint.java
```diff
@@ -0,0 +1,57 @@
...
+ * {@link org.springframework.pulsar.annotation.PulsarListenerConfigurer} to register
+ * endpoints programmatically.
+ *
+ * @param <T> Message payload type.
+ * @author Christophe Bornet
+ */
+public interface ReactivePulsarListenerEndpoint<T> extends GenericPulsarListenerEndpoint {
+
+	@Nullable
+	SubscriptionType getSubscriptionType();
```
onobc, line 40: Any reason that all of these fields were not moved to GPLE seeing as how they are in both PLE and RPLE?

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/AbstractReactivePulsarListenerEndpoint.java
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Base implementation for {@link ReactivePulsarListenerEndpoint}.
+ *
+ * @param <T> Message payload type.
+ * @author Christophe Bornet
+ */
+public abstract class AbstractReactivePulsarListenerEndpoint<T>
+		implements ReactivePulsarListenerEndpoint<T>, BeanFactoryAware, InitializingBean {
+
+	private final LogAccessor logger = new LogAccessor(LogFactory.getLog(getClass()));
```
onobc, line 55: 2 things:

* I don't think this is being used currently 
* there is also a `LogAccessor.<init>(Class)` which simplifies this a bit

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/AbstractReactivePulsarListenerEndpoint.java
```diff
@@ -0,0 +1,234 @@
...
+
+	@Override
+	public String getId() {
+		return this.id;
+	}
+
+	public void setTopics(String... topics) {
+		Assert.notNull(topics, "'topics' must not be null");
+		this.topics.clear();
+		this.topics.addAll(Arrays.asList(topics));
```
onobc, line 144: I doubt this will be done frequent enough to matter, but another option would be to just create a new immutable list on each invocation and replace the other one. I think this is fine as is though.

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerContainerFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * Factory for Pulsar message listener containers.
+ *
+ * @param <C> message listener container type.
+ * @param <E> listener endpoint type.
+ * @author Soby Chacko
+ * @author Christophe Bornet
+ */
+public interface GenericPulsarListenerContainerFactory<C, E extends GenericPulsarListenerEndpoint> {
+
+	C createListenerContainer(E endpoint);
```
onobc, line 29: I think javadoc would help on these 2 methods to understand the naming difference (one is a "listener" container - but they are both type `C`).
It is a bit confusing as-is.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/GenericPulsarListenerContainerRegistry.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * A registry for containers.
+ *
+ * @param <T> Message payload type.
+ * @author Christophe Bornet
+ */
+public interface GenericPulsarListenerContainerRegistry<T> {
+
+	@Nullable
+	T getListenerContainer(String id);
```
onobc, line 33: Do you mind adding javadoc here describing these methods? Also, what is the diff between "getListenerContainers" and "getAllListenerContainers"?

spring-pulsar/src/main/java/org/springframework/pulsar/listener/adapter/PulsarReactiveStreamingMessagingMessageListenerAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.springframework.pulsar.listener.reactive.ReactivePulsarMessageHandler;
+import org.springframework.pulsar.listener.reactive.ReactivePulsarStreamingHandler;
+
+import reactor.core.publisher.Flux;
+
+/**
+ * A {@link ReactivePulsarMessageHandler MessageListener} adapter that invokes a
+ * configurable {@link HandlerAdapter}; used when the factory is configured for the
+ * listener to receive individual messages.
```
onobc, line 32: I think this needs a bit of updating.

onobc (changes requested):
@cbornet great work!

I have left some minor comments. Rather than roundtrip this PR back and forth to get it "perfect" I am in favor of getting this PR in and then iterate subsequently as needed. There are enough moving pieces and this is the very 1st pass for the RPL. 

That being said, feel free to push any of my suggestions out to next PR etc.. I do think adding the javadoc on the interfaces would be helpful now though. I am requesting changes accordingly.

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpoint.java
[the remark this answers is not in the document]
cbornet: Sure

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerEndpointRegistry.java
cbornet: Ok

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListener.java
cbornet: OK. I guess the same can be said for the `PulsarListener::batch` javadoc ?

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/ReactivePulsarListenerEndpoint.java
cbornet: It was to keep reactive and imperative impls separated but it may not make sense to do it here.

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerContainerFactory.java
cbornet: I think this can be simplified if we introduce a `MessageListenerContainer extends Disposable, SmartlifeCycle` like in spring-kafka. Will give it a shot.

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListener.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/annotation/ReactivePulsarListener.java
cbornet: Done

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/ReactivePulsarListenerEndpoint.java
cbornet: Done

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/AbstractReactivePulsarListenerEndpoint.java
cbornet: Done

spring-pulsar/src/main/java/org/springframework/pulsar/config/GenericPulsarListenerContainerFactory.java
cbornet: Done.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/GenericPulsarListenerContainerRegistry.java
cbornet: Done

spring-pulsar/src/main/java/org/springframework/pulsar/listener/adapter/PulsarReactiveStreamingMessagingMessageListenerAdapter.java
cbornet: Done

spring-pulsar/src/main/java/org/springframework/pulsar/config/reactive/AbstractReactivePulsarListenerEndpoint.java
cbornet: Done

spring-pulsar/src/main/java/org/springframework/pulsar/config/ListenerEndpointRegistry.java
```diff
@@ -0,0 +1,257 @@
...
+ * @param <C> listener container type.
+ * @param <E> listener endpoint type.
+ * @author Soby Chacko
+ * @author Christophe Bornet
+ */
+public class ListenerEndpointRegistry<C extends MessageListenerContainer, E extends ListenerEndpoint<C>>
+		implements ListenerContainerRegistry, DisposableBean, SmartLifecycle, ApplicationContextAware,
+		ApplicationListener<ContextRefreshedEvent> {
+
+	private final Class<? extends C> type;
```
onobc, line 65: **[For Later]** A more descriptive name may improve readability where this is used below.

PR #211, opened 2022-11-17
Remove maxInFlight

cbornet:
There is no real use for this as the number of messages pulled is already capped by concurrency*32 (default prefetch of flatMap). 

maxInFlight may be removed from the reactive client


onobc:
LGTM. Thanks @cbornet

PR #215, opened 2022-11-17
Remove initialSubscriptionName from sender properties

cbornet:
`initialSubscriptionName` is an internal param used by the Pulsar client for DLT. It is not used in the reactive client and should be removed from `MutableReactiveMessageSenderSpec`


onobc:
@cbornet 👍🏻 LGTM

PR #224, opened 2022-11-19
Add more settings to Sender in PulsarReactiveProperties

cbornet:


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
```diff
@@ -146,11 +152,18 @@ public static class Sender {
...
+
 		/**
 		 * Maximum number of messages to be batched.
 		 */
 		private Integer batchingMaxMessages = 1000;
 
+		/**
+		 * Maximum number of bytes permitted in a batch.
+		 */
+		private DataSize batchingMaxBytes = DataSize.ofKilobytes(128);
```
onobc, line 165: **[UNRELATED_TO_PR_DIRECTLY]**
The config props (in general) are getting fully fleshed out now and some of the props can be further grouped (in this case the "batch" props are a candidate to put into a `Batch` sub-config props). I am not suggesting doing anything in this PR but it inspired me to create #225 where we can do this grouping when we do harmonizing across reactive/imperative config props. 

Please resolve this once digested - nothing to be done other than sharing info.

cbornet: Yes, very good idea 👍

onobc (approved):

PR #226, opened 2022-11-19
Add more settings to Consumer in PulsarReactiveProperties

cbornet:
Also removing `KeySharedMode` for now as `KeySharedPolicySticky` needs a param to configure `Ranges`  otherwise it fails. This can be added again later by adding support for ranges.

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
@@ -26,14 +26,13 @@ import org.apache.pulsar.client.api.CompressionType;
 import org.apache.pulsar.client.api.ConsumerCryptoFailureAction;
 import org.apache.pulsar.client.api.DeadLetterPolicy;
 import org.apache.pulsar.client.api.HashingScheme;
-import org.apache.pulsar.client.api.KeySharedMode;
-import org.apache.pulsar.client.api.KeySharedPolicy;
 import org.apache.pulsar.client.api.MessageRoutingMode;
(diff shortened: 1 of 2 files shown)


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
```diff
@@ -835,11 +834,8 @@ public ReactiveMessageConsumerSpec buildReactiveMessageConsumerSpec() {
 			map.from(this::getTopicsPattern).to(spec::setTopicsPattern);
 			map.from(this::getSubscriptionName).to(spec::setSubscriptionName);
 			map.from(this::getSubscriptionType).to(spec::setSubscriptionType);
-			map.from(this::getKeySharedMode).as((mode) -> switch (mode) {
-				case STICKY -> KeySharedPolicy.stickyHashRange();
-				case AUTO_SPLIT -> KeySharedPolicy.autoSplitHashRange();
-			}).to(spec::setKeySharedPolicy);
```
onobc, line 841: Yeh, I can envision something like the following. It would require a custom `Range` converter to take `1,100` -> Range(1, 100).

```yaml
spring.pulsar.reactive.consumer.key-shared:
  mode: sticky | auto
  ranges: 
    - 1,100
    - 101,199
    - 1000-1500
```

onobc (approved):
LGTM. Thanks @cbornet !

spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
cbornet: 👍

PR #227, opened 2022-11-20
Add KeySharedMode back to PulsarReactiveProperties

onobc:
The KeySharedMode property was removed (https://github.com/spring-projects-experimental/spring-pulsar/commit/d9018ea51c1818d31cb820a7f27909f9177cd3d3) as it required range(s) to be specified for STICKY mode. 

This adds ability to configure ranges and adds the KeySharedMode property back into PulsarReactiveProperties.

@cbornet PTAL

The `application.yml` looks like the following: 
```yaml
spring:
  pulsar:
    reactive:
      consumer:
        key-shared-mode: sticky
        key-shared-sticky-ranges:
          - start: 0
            end: 65535
```

And it just about the time of crafting the above yaml and playing around w/ the feature that I realize ranges defined globally like this are pretty useless. This is really something that each listener will need to define. Most likely via a customizer. 

Do you agree? 

This also applies to the [reader](https://github.com/spring-projects-experimental/spring-pulsar/blob/e97f9c6f94c123f56a58539662da63c354d59f1f/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java#L464).

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarReactiveProperties.java
@@ -30,6 +30,8 @@ import org.apache.pulsar.client.api.CompressionType;
 import org.apache.pulsar.client.api.ConsumerCryptoFailureAction;
 import org.apache.pulsar.client.api.DeadLetterPolicy;
 import org.apache.pulsar.client.api.HashingScheme;
+import org.apache.pulsar.client.api.KeySharedMode;
+import org.apache.pulsar.client.api.KeySharedPolicy;
 import org.apache.pulsar.client.api.MessageRoutingMode;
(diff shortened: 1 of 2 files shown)


cbornet:
Yes I agree. Hash ranges are need to be fine tuned for a given topic so they change for each listener.
That said, I might be completely wrong, but I think 99% of applications use will use only one listener.
And also 99.9% of applications won't use key shared sticky hash ranges.
So only having customization in the listener is probably sufficient for now and we can add something later in the properties if there's demand for it.

onobc:
> Yes I agree. Hash ranges are need to be fine tuned for a given topic so they change for each listener. That said, I might be completely wrong, but I think 99% of applications use will use only one listener. And also 99.9% of applications won't use key shared sticky hash ranges. So only having customization in the listener is probably sufficient for now and we can add something later in the properties if there's demand for it.

We should probably pull the property from the reader as well.

onobc:
Closing as per https://github.com/spring-projects-experimental/spring-pulsar/pull/227#issuecomment-1321872613

PR #229, opened 2022-11-21
Add more settings to Admin in PulsarProperties

cbornet:


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -1811,6 +1811,26 @@ public static class Admin {
 		 */
 		private Set<String> tlsProtocols;
 
+		/**
+		 * Connection time out.
+		 */
+		private Duration connectTimeout = Duration.ofMinutes(1);
```
onobc, line 1817: This will work as expected once https://github.com/apache/pulsar/pull/17375 gets merged. We ran into this originally in https://github.com/apache/pulsar/issues/17333

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -61,12 +64,28 @@
 	 * @param adminConfig the {@link PulsarAdmin} configuration
 	 */
 	public PulsarAdministration(Map<String, Object> adminConfig) {
-		this.adminBuilder = PulsarAdmin.builder().loadConf(adminConfig);
+
+		this.adminBuilder = PulsarAdmin.builder();
+
+		Map<String, Object> conf = new HashMap<>(adminConfig);
```
onobc, line 70: Wdyt about abstracting this via a utility method like we do in the other client builders?

onobc (changes requested):
@cbornet LGTM

One possible improvement comment - feel free to do or not.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
cbornet: done

onobc (approved):

PR #233, opened 2022-11-21
Update with new MessageResult::acknowledge(Message) function

cbornet:


onobc (approved):
Nice improvement in the client! LGTM.

PR #241, opened 2022-11-27
Update to Spring Boot 3.0.0

onobc:
- Also update Micrometer docs gen to M7

### DIFF
--- a/spring-pulsar-docs/build.gradle
+++ b/spring-pulsar-docs/build.gradle
@@ -6,7 +6,7 @@ plugins {
 description = 'Spring Pulsar Docs'
 
 ext {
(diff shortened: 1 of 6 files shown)


spring-pulsar-docs/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 description = 'Spring Pulsar Docs'
 
 ext {
-	micrometerDocsVersion="1.0.0-SNAPSHOT"
+	micrometerDocsVersion="1.0.0-M7"
```
onobc, line 9: The MM docs generator has actually moved into a single module under `1.0.0`. I will update this is in a follow on PR though.

PR #244, opened 2022-11-27
Change autoCertRefreshTimeMs to autoCertRefreshSeconds

cbornet:
The 2.12 Pulsar client will handle the autoCertRefreshTime in CCD but the property will be named `autoCertRefreshSeconds`


onobc (approved):
@cbornet - you are too AWESOME. Based on your previous review I had just fetched latest from main and was about to `git checkout -b cbono-rename-autocertrefresh` and then looked here only to find you already followed suit and renamed it for us. Thank you my friend.

cbornet:
You're welcome 😉

PR #246, opened 2022-11-28
Remove custom argument resolvers

sobychacko:
- Remove custom argument resolvers from PulsarListenerEndpointRegistrar, as this is an un-used feature at the moment.

- The default HandlerMethodArgumentResolver is capable of adapting the all the method arguments in all the PulsarListener use cases we encountered so far. We will add it back if custom argument resolution is necessary.


onobc (approved):
I like this simplification @sobychacko . The fact that there are no tests to remove also tells me that we did not have test coverage for this feature anyways. As you said earlier, we can always bring back when/if there is a user feature request or other internal need.

PR #247, opened 2022-11-29
Remove custom MessageHandlerMethodFactory

sobychacko:
- Remove custom MessageHandlerMethodFactory from PulsarListenerEndpointRegistrar
- We will reintroduce it, if there is a valid use case that requires it


onobc (approved):
LGTM @sobychacko - thanks for the cleanup.

Issue #163, opened 2022-10-16
[Build] Fix CI deprecation warnings


onobc:
Each of our CI runs has notifications/warnings about several deprecated features we are using in our workflows.

List of deprecations:  https://github.com/spring-projects-experimental/spring-pulsar/actions/runs/3238804365

onobc:
This has been completed via various commits.

PR #260, opened 2022-12-16
Fix CI deprecation warnings

wangqinggo:
#163


onobc:
Hi @wangqinggo , 

Thank you for the contribution but we prefer to handle dependencies and the CI/build internally. The reason is they have the potential to cause unforeseen issues and can be risky. 

I also realized that we had not stated that anywhere in the docs yet. I have corrected that by adding a pull request template and issue template to with these details included. 

Again, thank you for taking the time. Please feel free to find other issues in the backlog to work on. Soon we will add labels to indicate issues that are `ideal for contribution`.

PR #262, opened 2022-12-16
Initial foundation for SCSt binder

sobychacko:
- Mostly layout


spring-pulsar-dependencies/build.gradle
```diff
@@ -13,6 +13,7 @@ ext {
 	pulsarVersion = '2.10.2'
 	pulsarClientReactiveVersion = '0.1.0'
 	springBootVersion = '3.0.0'
+	springCloudStreamVersion = '4.0.0-SNAPSHOT'
```
onobc, line 16: Everything else is locked on full GA versions currently. I think we should make them all snapshot or all GA for consistency. I think we locked onto GA just before we released. I understand the benefit of being on snapshots and I also understand the benefits of not being on snapshots :) 

However, we should choose the mode that the rest of the projects use. What does SK do?

spring-pulsar-spring-cloud-stream-binder/build.gradle
```diff
@@ -0,0 +1,14 @@
+plugins {
+    id 'org.springframework.pulsar.spring-module'
+}
+
+description = 'Spring Cloud Stream Binder for Apache Pulsar'
+
+dependencies {
+    api project (':spring-pulsar')
+    api('org.springframework.cloud:spring-cloud-stream') {
+        exclude group: 'javax.activation', module: 'javax.activation-api'
```
onobc, line 10: How did you know to exclude these? Is there any other baggage we want to exclude?

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinder.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.cloud.stream.provisioning.ProducerDestination;
+import org.springframework.integration.core.MessageProducer;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.pulsar.spring.cloud.stream.binder.properties.PulsarConsumerProperties;
+import org.springframework.pulsar.spring.cloud.stream.binder.properties.PulsarProducerProperties;
+import org.springframework.pulsar.spring.cloud.stream.binder.provisioning.PulsarTopicProvisioner;
+
+/**
+ * {@link org.springframework.cloud.stream.binder.Binder} implementation for Apache
```
onobc, line 34: **[NIT]** I would prefer the short name w/ an import.

spring-pulsar-spring-cloud-stream-binder/build.gradle
sobychacko: Build complained.

spring-pulsar-dependencies/build.gradle
sobychacko: Ya, now that we have GA, we can use that for a while (`4.0.0` that is).

PR #276, opened 2023-01-19
Spring Cloud Stream Pulsar Binder

sobychacko:
Basic consumer and producer bindings.
Sample app (end to end)


spring-pulsar-sample-apps/sample-pulsar-binder/src/main/java/org/springframework/pulsar/sample/binder/SpringPulsarBinderSampleApp.java
```diff
@@ -0,0 +1,48 @@
...
+public class SpringPulsarBinderSampleApp {
+
+	private final Logger logger = LoggerFactory.getLogger(SpringPulsarBinderSampleApp.class);
+
+	public static void main(String[] args) {
+		SpringApplication.run(SpringPulsarBinderSampleApp.class, args);
+	}
+
+	@Bean
+	public Supplier<String> supplier() {
```
onobc, line 39: I think using `supplier` and `consumer` as the function names can be a bit confusing when looking at the properties as those are generic terms that exist in other areas of the properties etc.. Wdyt about renaming something like `timeSupplier` and `timeLogger`?

spring-pulsar-sample-apps/sample-pulsar-binder/src/main/resources/application.yml
```diff
@@ -0,0 +1,16 @@
...
+        consumer-in-0:
+          destination: supplier-out-0
+          consumer:
+            use-native-decoding: true
+      pulsar:
+        bindings:
+          consumer-in-0:
+            consumer:
+              subscription-name: my-scst-sub1
+              schema-type: STRING
```
onobc, line 16: I see you did not need to set schema type on producer. Is that because the payload is always `Object` and then it defaults to `Schema.BYTES` in that case?

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarBinderConfigurationProperties.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.spring.cloud.stream.binder.properties;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "spring.cloud.stream.pulsar.binder")
+public class PulsarBinderConfigurationProperties {
```
onobc, line 22: I am guessing this is just not used yet, but will be?

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/config/PulsarBinderConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.pulsar.autoconfigure.PulsarProperties;
+import org.springframework.pulsar.core.PulsarConsumerFactory;
+import org.springframework.pulsar.core.PulsarTemplate;
+import org.springframework.pulsar.spring.cloud.stream.binder.PulsarMessageChannelBinder;
+import org.springframework.pulsar.spring.cloud.stream.binder.properties.PulsarExtendedBindingProperties;
+import org.springframework.pulsar.spring.cloud.stream.binder.provisioning.PulsarTopicProvisioner;
+
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnMissingBean(Binder.class)
```
onobc, line 32: Does this work since this is not really an `@Autoconfiguration` but rather a config that gets loaded by SCSt binder factory?

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinder.java
```diff
@@ -40,31 +54,76 @@ public class PulsarMessageChannelBinder extends
...
+		}
+		pulsarContainerProperties.setSubscriptionName(properties.getExtension().getSubscriptionName());
+		DefaultPulsarMessageListenerContainer<?> container = new DefaultPulsarMessageListenerContainer<>(
+				this.pulsarConsumerFactory, pulsarContainerProperties);
+		pulsarMessageDrivenChannelAdapter.setMessageListenerContainer(container);
+		return pulsarMessageDrivenChannelAdapter;
+	}
+
+	// This is just a place-holder impl
+	public Schema<?> toSchema(SchemaType schemaType) {
```
onobc, line 111: Got it - I see what you are doing here. I will add the API to the resolver in the next day or so.

onobc (approved):
@sobychacko looking good. 

I have comments inline (mostly of the question variety). Any of them that are "asks" to change, please just handle in subsequent PRs as we want to get this one merged now.



### Out of Band comment

I did not see the Provisioner in this PR because it is already in the base impl, but I think switching the `PulsarDestination/PulsarProducerDestination/PulsarConsumerDestination` to use records would simplify things a bit.

sobychacko:
> 
We will convert them to records in the subsequent PR's.

Issue #269, opened 2023-01-05
Allow user to configure mappings from message type to topic name and schema 


onobc:
Consider the following code[^1]: 

```java
@Autowired PulsarTemplate<Type1> template1;
@Autowired PulsarTemplate<Type2> template2;

void send1(...) {
  template1.setSchema(Schema.JSON(Type1));
  template1.send("topic-1", new Type1());
}

void send2(...) {
  template2.setSchema(Schema.JSON(Type2));
  template2.send("topic-2", new Type2());
}
```

It would be convenient for the user to be able to configure a mapping between message types to schema and topics to use. Something like:
```yaml
spring.pulsar.producer.topic-mappings:
  acme.Type1: "topic1"
  acme.Type2: "topic2"

spring.pulsar.producer.schema-mappings:
  acme.Type1: SchemaType.JSON(acme.Type1)
  acme.Type2: SchemaType.JSON(acme.Type2)
```

This would reduce the boilerplate - such as:
```java
@Autowired PulsarTemplate<Type1> template1;
@Autowired PulsarTemplate<Type2> template2;

void send1(...) {
  template1.send(new Type1());
}

void send2(...) {
  template2.send(new Type2());
}
```

[^1]: https://stackoverflow.com/questions/75023746/defining-multiple-pulsar-producers-in-spring-boot-application

gootik:
Some extra information on my setup right now. So far I've managed to reduce a little bit of the repetition by creating specific PulsarTemplate beans. This at least solves the schema issue, but I'm not entirely sure how this will play with caching. I believe this will be fine since only one `CachingPulsarProducerFactory` is auto generated. 

But this can't be taken further since we can't force the factory to use a specific topic.

```
@Configuration
public class MessageQueueConfig {
  @Bean
  public PulsarTemplate<Type1> createAccountActivityTemplate(PulsarProducerFactory<Type1> factory) {

    var template = new PulsarTemplate<>(factory);
    template.setSchema(Schema.JSON(Type1.class));

    return template;
  }

  @Bean
  public PulsarTemplate<Type2> createAccountActivityTemplate(PulsarProducerFactory<Type2> factory) {

    var template = new PulsarTemplate<>(factory);
    template.setSchema(Schema.JSON(Type2.class));

    return template;
  }
```

onobc:
Hi @gootik ,
The topic-side of the equation is in main branch now as well. 

[...]

PR #280, opened 2023-01-20
Listener endpoints use schema resolver

onobc:
Removes the schema mapping code out of the method listener endpoints and into the SchemaResolver. 

See #269

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
@@ -17,6 +17,8 @@
 package org.springframework.pulsar.core;
 
 import java.nio.ByteBuffer;
+import java.sql.Time;
+import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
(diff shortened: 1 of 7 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
```diff
@@ -103,7 +116,64 @@ private Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean r
 		return this.customSchemaMappings.getOrDefault(messageClass, (returnDefault ? Schema.BYTES : null));
 	}
 
-	@SuppressWarnings({ "unchecked" })
+	@Override
+	@SuppressWarnings("unchecked")
+	public <T> Schema<T> getSchema(SchemaType schemaType, @Nullable ResolvableType messageType) {
```
onobc, line 121: @sobychacko you can use this API in the binder and pass in null for `messsageType` which will give mappings for all of the primitives. 

```java
	public Schema<?> toSchema(SchemaType schemaType) {
		SchemaResolver schemaResolver = new DefaultSchemaResolver();
		return schemaResolver.getSchema(schemaType, null);
	}
```

Then just need to make it possible for users to wire in custom mappings.

PR #282, opened 2023-01-20
Pulsar binder schema related changes

sobychacko:

### DIFF
--- a/spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarBinderIntegrationTests.java
+++ b/spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarBinderIntegrationTests.java
@@ -46,13 +46,46 @@ class PulsarBinderIntegrationTests implements PulsarTestContainerSupport {
 	void basicProducerConsumerBindingEndToEnd(CapturedOutput output) {
 		SpringApplication app = new SpringApplication(BasicScenarioConfig.class);
 		app.setWebApplicationType(WebApplicationType.NONE);
-		try (ConfigurableApplicationContext context = app.run(
+		try (ConfigurableApplicationContext ignored = app.run(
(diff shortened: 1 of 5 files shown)


spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarBinderIntegrationTests.java
```diff
@@ -56,6 +56,39 @@ void basicProducerConsumerBindingEndToEnd(CapturedOutput output) {
...
+				"--spring.cloud.stream.pulsar.bindings.piLogger-in-0.consumer.schema-type=FLOAT",
+				"--spring.cloud.stream.pulsar.bindings.piSupplier-out-0.producer.schema-type=FLOAT",
+				"--spring.cloud.stream.pulsar.bindings.piLogger-in-0.consumer.subscription-name=native-encoding-decoding-sub-1")) {
+			Awaitility.await().atMost(Duration.ofSeconds(10))
+					.until(() -> output.toString().contains("Hello binder: 3.14"));
+		}
+	}
+
+	@Test
+	void springCloudStreamConversionForNonTextTypes(CapturedOutput output) {
```
onobc, line 78: This is the same as the `basicProducerConsumerBindingEndToEnd` except that it uses `Float` instead of `String` values in the functions, correct? If so, maybe we can name them similarly? It took me a sec to figure out the diff between these 3 tests etc..

onobc (approved):
LGTM

spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarBinderIntegrationTests.java
sobychacko: Ya, I wanted some tests with non-text payloads doing SCSt conversion.

PR #285, opened 2023-01-22
Add schema resolution to binder

onobc:


spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinder.java
```diff
@@ -118,6 +129,40 @@ protected MessageProducer createConsumerEndpoint(ConsumerDestination destination
...
+			}
+			else if (messageKeyType != null && messageValueType != null) {
+				resolvableType = ResolvableType.forClassWithGenerics(KeyValue.class, messageKeyType, messageValueType);
+			}
+			if (resolvableType == null) {
+				throw new IllegalArgumentException(
+						"'message-type' OR ('message-key-type' AND 'message-value-type') required for 'schema-type' NONE");
+			}
+		}
+		// TODO if schema == null then default lookup bean Schema<?> w/ name == binding
```
onobc, line 162: This would be a nice fallback to add so that users can define a `Schema<?>` bean named `schemaFor<binding-name>` in case they need to provide a schema that we don't support w/ the above flow.

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinder.java
```diff
@@ -118,6 +129,40 @@ protected MessageProducer createConsumerEndpoint(ConsumerDestination destination
...
+			@Nullable Class<?> messageKeyType, @Nullable Class<?> messageValueType) {
+		if (schemaType == null) {
+			schemaType = SchemaType.NONE;
+		}
+		ResolvableType resolvableType = null;
+		if (schemaType.isStruct()) {
+			resolvableType = ResolvableType.forClass(Objects.requireNonNull(messageType,
+					"'message-type' required for 'schema-type' " + schemaType.name()));
+		}
+		else if (schemaType == SchemaType.KEY_VALUE) {
```
onobc, line 144: There is a caveat w/ the `KEY_VALUE` schema resolution currently (not just here - same for template and listener)...

Complex types will map to `BYTES` unless the user adds a custom mapping for the type.  For example, given a `KevValue<String, Foo>` - it will use `Schema.STRING` and `Schema.BYTES`. 

However, if we add a custom mapping to the resolver for `Foo.class -> JSONSchema.of(Foo.class)` then it will map properly.

Issue #268, opened 2023-01-05
PulsarTemplate setSchema is not thread-safe


onobc:
Consider the following code[^1]: 
```java
@Autowired PulsarTemplate<Type1> template1;
@Autowired PulsarTemplate<Type2> template2;

void send1(...) {
  template1.setSchema(Schema.JSON(Type1));
  template1.send("topic-1", ...);
}

void send2(...) {
  template2.setSchema(Schema.JSON(Type2));
  template2.send("topic-2", ...);
}
```

Because there is only a single template instance that is assigned to both `template1` and template2` it is possible that the wrong schema could be used on the wrong data type. 

* **t1** `send1` called and set schema to type1 and then thread gets pre-empted
* **t2** `send2` called and set schema to type2 and then gets pre-empted
* **t3** `send1` continues and does a send w/ a type1 object
* 💥 type1 object does not work well w/ type2 schema

## Suggested Fix
Move the `setSchema` to the send API params

[^1]: https://stackoverflow.com/questions/75023746/defining-multiple-pulsar-producers-in-spring-boot-application

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/ed2a67b0c4b8b981362dec76e178e6ccb2b14327

PR #287, opened 2023-01-25
Remove setSchema API from templates

onobc:
For both PulsarTemplate and ReactivePulsarTemplate:

* remove setSchema()
* add schema param in simple API
* add schema param in fluent API

Update all tests and samples accordingly

See #268

### DIFF
--- a/spring-pulsar-docs/src/main/asciidoc/schema-info/schema-info-template.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/schema-info/schema-info-template.adoc
@@ -1,19 +1,12 @@
 == Specifying Schema Information
 If you use Java primitive types, the framework auto-detects the schema for you, and you need not specify any schema types for publishing the data.
-However, if you use any complex types (such as `JSON`, `AVRO`, `PROTOBUF`, and others), you need to set the proper schema type on the `{template-class}` before invoking any send operations, as the following example shows for JSON:
-
-====
-[source, java]
-----
-template.setSchema(Schema.JSON(Foo.class));
-----
-====
+However, if you use any complex types (such as `JSON`, `AVRO`, `PROTOBUF`, and others), you need to specify the proper schema when invoking send operations on the `{template-class}`.
 
 IMPORTANT: Complex Schema types that are currently supported are JSON, AVRO, PROTOBUF, and KEY_VALUE w/ INLINE encoding.
 
 === Custom Schema Mapping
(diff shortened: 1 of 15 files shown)


spring-pulsar-docs/src/main/asciidoc/schema-info/schema-info-template.adoc
```diff
@@ -1,19 +1,12 @@
...
-====
+However, if you use any complex types (such as `JSON`, `AVRO`, `PROTOBUF`, and others), you need to specify the proper schema when invoking send operations on the `{template-class}`.
 
 IMPORTANT: Complex Schema types that are currently supported are JSON, AVRO, PROTOBUF, and KEY_VALUE w/ INLINE encoding.
 
 === Custom Schema Mapping
-As an alternative to specifying the schema on the `{template-class}` for complex types, the schema resolver can be configured with mappings for the types.
-This removes the need to set the schema on the template as the framework consults the resolver using the outgoing message type.
+As an alternative to specifying the schema when invoking send operations on the `{template-class}` for complex types, the schema resolver can be configured with mappings for the types.
+This removes the need to set the specify the schema as the framework consults the resolver using the outgoing message type.
```
sobychacko, line 9: `need to set specifying the schema`?

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
```diff
@@ -34,7 +34,7 @@
 import reactor.core.publisher.Mono;
 
 /**
- * A thread-safe template for executing high-level reactive Pulsar operations.
+ * A template for executing high-level reactive Pulsar operations.
```
sobychacko, line 37: why did we remove the thread-safe part from the comment?

sobychacko (changes requested):

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
onobc: I removed it because it was a holdover from the Kafka template and we don't really distinguish this elsewhere. 

If we want to bring it back, I would propose that we use @ThreadSafe and also do it for all classes that fit that bill rather than just the template.

PR #290, opened 2023-01-26
Create partitioned topics once per ns

lanwen:
Thank you for the nice module, however, I faced an issue with the Administration behaviour regarding the topic creation in several namespaces.

If we define separate topics in different namespaces, the administration shouldn't try to create each topic several times. 
This PR changes the `topicsToCreate` scope to be at the same level where the creation happens. 

Otherwise, it creates the topic in one namespace, adds a topic in another namespace to the same set and without cleaning it up tries to create 2 topics, which leads to 409 response from pulsar.

This PR adds a test to demonstrate the case.


onobc:
Hi @lanwen

Thank you for finding the issue and even more for submitting the fix! 👍🏻 . We will take a review asap.

onobc:
Hi @lanwen ,

Thanks again for the contribution. Nice work. 

I added a small polish commit on top of yours that:

- ran the "./gradlew format"
- add you as @author 
- re-ordered new test utility methods

We look forward to more contributions and seeing you in the Pulsar community. 

P.S. If you like the framework so far, would you so kindly star the repo?

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/392cb6f5a5645fb961ee5ecdd41ed775dfa372ae

Issue #286, opened 2023-01-24
Add ability to pause/resume consumer


pkokoshnikov:
I want to be able in some situations to pause consuming. For example if service responds for example 500, it would be better to stop to do requests. Wait for example 10s and try again. If one request is ok then we can resume consuming.

sobychacko:
@pkokoshnikov That certainly sounds like a very interesting use case. We will look into it as a new feature on the container side that the consumer uses.

sobychacko:
@pkokoshnikov We added the pause/resume feature. Please look at the latest commit and the [documentation](https://docs.spring.io/spring-pulsar/docs/current-SNAPSHOT/reference/html/#_pausing_and_resuming_message_listener_containers). 

If you like Spring Pulsar so far, please feel free to [star](https://github.com/spring-projects-experimental/spring-pulsar/stargazers) the repo:-) And more importantly, continue to give us feedback (or feature requests) so that we can improve things further. You can also contribute any bug fixes or features in the form of pull requests.

PR #292, opened 2023-01-28
Add the ability to pause/resume container

sobychacko:
Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/286

### DIFF
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -1146,7 +1146,7 @@ void listen1(String message) {
 ----
 ====
 
-TIP: The id parameter is the value of the @PulsarListener id attribute (when pausing a @PulsarListener).
+TIP: The id parameter is the value of the `@PulsarListener` id attribute (when pausing a `@PulsarListener`).
 
 When running the above app, you will see the following output:
 
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -1100,6 +1100,70 @@ On retry, the container sends a new batch of messages, starting with the failed 
 If it fails again, it is retried, until the retries are exhausted, at which point the message is sent to the DLT.
 At that point, the message is acknowledged by the container, and the listener is handed over with the subsequent messages in the original batch.
 
+=== Pausing and Resuming Message Listener Containers
+
+There are situations in which an application might want to pause message consumption temporarily and then resume later.
+Spring for Apache Pulsar provides the ability to pause and resume the underlying message listener containers.
+When the Pulsar message listener container is paused, any polling done by the container to receive data from the Pulsar consumer will be paused.
+Similarly, when the container is resumed, the next poll starts returning data if the topic has any new records added while paused.
+The container provided pause and resume features sit a level atop of the similar capabilities provided by the Pulsar consumer which means that in addition to pausing and resuming the Pulsar consumer, the container will not even invoke the `receive` method on the consumer while it is paused.
+
+In order for applications to pause and resume the listener container, they need to inject a special bean called `PulsarListenerEndpointRegistry`.
+Once the application gets a handle on this bean, then it can query for the listener container by its id and then invoke pause and resume operations.
+
+Here is an example.
+
+====
+[source, java]
+----
+@Bean
+ApplicationRunner runner1(PulsarTemplate<String> pulsarTemplate, PulsarListenerEndpointRegistry   pulsarListenerEndpointRegistry) {
+
+    String topic = "hello-pulsar";
+
+    return args -> {
+        for (int i = 0; i < 5; i++) {
+            pulsarTemplate.send(topic, "This is message " + (i + 1));
+        };
+        Thread.sleep(10_000);
+        PulsarMessageListenerContainer listenerContainer = pulsarListenerEndpointRegistry.getListenerContainer("consumer-pause-resume-listener");
+        System.out.println("Pausing and sleeping for 10 seconds");
+        listenerContainer.pause();
+        Thread.sleep(10_000);
+        for (int i = 5; i < 10; i++) {
+            pulsarTemplate.send(topic, "This is message " + (i + 1));
+        }
+        Thread.sleep(10_000);
+        System.out.println("Slept for 10 seconds and resuming");
+        listenerContainer.resume();
+    };
+}
+
+@PulsarListener(id = "consumer-pause-resume-listener", subscriptionName = "my-subscription", topics = "hello-pulsar")
+void listen1(String message) {
+    System.out.println(message);
+}
+
+----
+====
+
+When running this app, you will see the following output:
+
+```
+This is message 1
+This is message 2
+This is message 3
+This is message 4
+This is message 5
+Pausing and sleeping for 10 seconds
+Slept for 10 seconds and resuming
+This is message 6
+This is message 7
+This is message 8
+This is message 9
+This is message 10
+```
+
 == Publishing and Consuming Partitioned Topics
 
 In the following example, we publish to a topic called `hello-pulsar-partitioned`.
--- a/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
@@ -1106,10 +1106,9 @@ There are situations in which an application might want to pause message consump
 Spring for Apache Pulsar provides the ability to pause and resume the underlying message listener containers.
 When the Pulsar message listener container is paused, any polling done by the container to receive data from the Pulsar consumer will be paused.
 Similarly, when the container is resumed, the next poll starts returning data if the topic has any new records added while paused.
-The container provided pause and resume features sit a level atop of the similar capabilities provided by the Pulsar consumer which means that in addition to pausing and resuming the Pulsar consumer, the container will not even invoke the `receive` method on the consumer while it is paused.
 
 In order for applications to pause and resume the listener container, they need to inject a special bean called `PulsarListenerEndpointRegistry`.
-Once the application gets a handle on this bean, then it can query for the listener container by its id and then invoke pause and resume operations.
+To pause or resume a listener container, first obtain the container instance by querying the PulsarListenerEndpointRegistry bean and then invoke the pause/resume API on the instance.
 
 Here is an example.
 
@@ -1147,7 +1146,9 @@ void listen1(String message) {
 ----
 ====
 
-When running this app, you will see the following output:
+TIP: The id parameter is the value of the @PulsarListener id attribute (when pausing a @PulsarListener).
+
+When running the above app, you will see the following output:
 
 ```
 This is message 1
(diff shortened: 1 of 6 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -602,6 +622,18 @@ private void handleAck(Message<T> message) {
 			AbstractAcknowledgement.handleAckByMessageId(this.consumer, message.getMessageId());
 		}
 
+		public void pause() {
```
onobc, line 625: As we have `pause/resume` in several places, is there already a Spring abstraction for it? If not, should we create a `Pauseable` interface and use accordingly?

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -191,6 +191,24 @@ private void publishConsumerFailedToStart() {
...
+		if (consumer != null) {
+			consumer.pause();
+		}
+	}
+
+	@Override
+	public void resume() {
+		super.resume();
+		DefaultPulsarMessageListenerContainer<T>.Listener consumer = this.listenerConsumer;
+		if (consumer != null) {
```
onobc, line 207: I think we will want to resume the consumer before we call super.resume which signals to read from the consumer. As it is, I could imagine a race condition where the thread gets pre-empted at line 208 but the container resumes and tries to hit the consumer for data but its not yet resumed.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -955,4 +957,53 @@ void pulsarMessagesBatchListenerWithHeaders(Messages<String> data,
...
 	}
 
+	@Nested
+	@ContextConfiguration(classes = ConsumerPauseTest.ConsumerPauseConfig.class)
+	class ConsumerPauseTest {
+
+		private static final CountDownLatch latch = new CountDownLatch(10);
+
+		@Autowired
+		PulsarListenerEndpointRegistry pulsarListenerEndpointRegistry;
```
onobc, line 967: **[NIT]** Can this be private?

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -955,4 +957,53 @@ void pulsarMessagesBatchListenerWithHeaders(Messages<String> data,
...
+			// All latch must be received by now
+			assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
+		}
+
+		@EnablePulsar
+		@Configuration
+		static class ConsumerPauseConfig {
+
+			@PulsarListener(id = "consumerPauseListener", subscriptionName = "consumer-pause-subscription",
+					topics = "consumer-pause-topic", properties = { "receiverQueueSize=1" })
```
onobc, line 1000: Do we need to mention anything about the `receiverQueueSize` in the docs? I think because we control the state in the container, we do not have to enforce this. I am guessing you are just doing that here to aid in the test?

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -955,4 +957,53 @@ void pulsarMessagesBatchListenerWithHeaders(Messages<String> data,
...
+		@Test
+		void containerPauseAndResumeSuccessfully() throws Exception {
+			for (int i = 0; i < 3; i++) {
+				pulsarTemplate.send("consumer-pause-topic", "hello-" + i);
+			}
+			// wait until all 3 messages are received by the listener
+			Awaitility.await().timeout(Duration.ofSeconds(10)).until(() -> latch.getCount() == 7);
+			PulsarMessageListenerContainer container = pulsarListenerEndpointRegistry
+					.getListenerContainer("consumerPauseListener");
+			if (container != null) {
```
onobc, line 978: We should probably assertThat it is not null prior to this line.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
```diff
@@ -955,4 +957,53 @@ void pulsarMessagesBatchListenerWithHeaders(Messages<String> data,
...
+	@ContextConfiguration(classes = ConsumerPauseTest.ConsumerPauseConfig.class)
+	class ConsumerPauseTest {
+
+		private static final CountDownLatch latch = new CountDownLatch(10);
+
+		@Autowired
+		PulsarListenerEndpointRegistry pulsarListenerEndpointRegistry;
+
+		@Test
+		void containerPauseAndResumeSuccessfully() throws Exception {
```
onobc, line 970: Great test! It may be helpful to add a test that continues to put data on the topic while the container is paused (another thread) with a large `receiverQueueSize` in order to ensure it does what we expect in that case.

onobc (changes requested):
@sobychacko this is looking good. Most of comments are nits/docs related. There are couple of bigger questions though.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -338,7 +356,9 @@ public void run() {
 				try {
 					if (!inRetryMode.get() && !messagesPendingInBatch.get()) {
 						DefaultPulsarMessageListenerContainer.this.receiveInProgress.set(true);
-						messages = this.consumer.batchReceive();
+						if (!isPaused()) {
```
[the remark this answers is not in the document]
sobychacko: I will look into it and add this in a subsequent PR.

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -1100,6 +1100,70 @@ On retry, the container sends a new batch of messages, starting with the failed
 If it fails again, it is retried, until the retries are exhausted, at which point the message is sent to the DLT.
 At that point, the message is acknowledged by the container, and the listener is handed over with the subsequent messages in the original batch.
 
+=== Pausing and Resuming Message Listener Containers
+
+There are situations in which an application might want to pause message consumption temporarily and then resume later.
+Spring for Apache Pulsar provides the ability to pause and resume the underlying message listener containers.
+When the Pulsar message listener container is paused, any polling done by the container to receive data from the Pulsar consumer will be paused.
+Similarly, when the container is resumed, the next poll starts returning data if the topic has any new records added while paused.
+The container provided pause and resume features sit a level atop of the similar capabilities provided by the Pulsar consumer which means that in addition to pausing and resuming the Pulsar consumer, the container will not even invoke the `receive` method on the consumer while it is paused.
```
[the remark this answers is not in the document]
sobychacko: Sure, good idea.

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -1100,6 +1100,70 @@ On retry, the container sends a new batch of messages, starting with the failed
...
+In order for applications to pause and resume the listener container, they need to inject a special bean called `PulsarListenerEndpointRegistry`.
+Once the application gets a handle on this bean, then it can query for the listener container by its id and then invoke pause and resume operations.
+
+Here is an example.
+
+====
+[source, java]
+----
+@Bean
+ApplicationRunner runner1(PulsarTemplate<String> pulsarTemplate, PulsarListenerEndpointRegistry   pulsarListenerEndpointRegistry) {
```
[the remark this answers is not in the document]
sobychacko: TIP is a good idea, I will add it along with the extraction of the registry. I think having a complete example like the way it is there doesn't hurt and might give some extra use-case perspectives.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/AbstractPulsarMessageListenerContainer.java
```diff
@@ -223,4 +225,18 @@ public void setPulsarConsumerErrorHandler(PulsarConsumerErrorHandler pulsarConsu
 		this.pulsarConsumerErrorHandler = pulsarConsumerErrorHandler;
 	}
 
+	@Override
```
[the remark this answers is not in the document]
sobychacko: https://github.com/spring-projects-experimental/spring-pulsar/issues/296

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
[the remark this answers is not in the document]
sobychacko: https://github.com/spring-projects-experimental/spring-pulsar/issues/297

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Let's do that outside the scope of this PR.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
sobychacko: Good point to add this to the docs (just as a passing reference).

sobychacko: But i will do that in a subsequent PR though.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerTests.java
sobychacko: ok

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
```diff
@@ -1147,7 +1146,9 @@ void listen1(String message) {
 ----
 ====
 
-When running this app, you will see the following output:
+TIP: The id parameter is the value of the @PulsarListener id attribute (when pausing a @PulsarListener).
```
onobc, line 1149: **[TinyNit]** Don't hate me, but do you mind formatting `@PulsarListener` w/ "` `"?

spring-pulsar-docs/src/main/asciidoc/pulsar.adoc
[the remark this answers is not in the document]
onobc: The problem w/ the example is that it may quickly become stale as it is not compiled code. Also it is using `System.out` and has some other things that would not pass the code formatter etc. I think the 2nd snippet that you have added is sufficient to show how to get the container from the registry and operate on it. The pause/resume is an understandable operation that probably does not need more than that.

onobc (approved):

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/53bd54af4cd499bf1cc88d96a5a2de1ff5991e5e

PR #293, opened 2023-01-28
Add topic resolver

onobc:
- Topic resolver with configurable message type to topic name mappings

See #269


sobychacko:
Great PR! merged upstream.

PR #294, opened 2023-01-28
Update SCSt binder sample

onobc:
Use 2 different types of schema now that the setSchema is thread-safe.

### DIFF
--- a/spring-pulsar-sample-apps/sample-pulsar-binder/src/main/java/org/springframework/pulsar/sample/binder/SpringPulsarBinderSampleApp.java
+++ b/spring-pulsar-sample-apps/sample-pulsar-binder/src/main/java/org/springframework/pulsar/sample/binder/SpringPulsarBinderSampleApp.java
@@ -37,21 +37,28 @@ public class SpringPulsarBinderSampleApp {
 	}
 
 	@Bean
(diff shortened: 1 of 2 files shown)


spring-pulsar-sample-apps/sample-pulsar-binder/src/main/java/org/springframework/pulsar/sample/binder/SpringPulsarBinderSampleApp.java
```diff
@@ -37,21 +37,28 @@ public static void main(String[] args) {
...
-	public Supplier<String> timeSupplier() {
-		return () -> String.valueOf(System.currentTimeMillis());
+	public Supplier<Time> timeSupplier() {
+		return () -> new Time(String.valueOf(System.currentTimeMillis()));
 	}
 
 	@Bean
-	public Function<String, EnhancedTime> timeProcessor() {
-		return (rawTime) -> new EnhancedTime(rawTime, "5150");
+	public Function<Time, EnhancedTime> timeProcessor() {
```
onobc, line 45: Before #268 this was not possible because both bindings would attempt to call `setSchema` and things would break. All good now though.

Issue #296, opened 2023-01-30
Pause/Resume improvements


sobychacko:
I think it would be more consistent and easier to follow if we instead do pause/resume exactly how start/stop is currently implemented. This would also remove the need to attempt to use the lifecycleMonitor in the subclass. 

Basically, we would have an abstract `doPause()/doResume()` where the concurrent does the 1-N delegation call to its child containers and each child container then does the work to pause/resume (as you have it now).

_Originally posted by @onobc in https://github.com/spring-projects-experimental/spring-pulsar/pull/292#discussion_r1089827965_

onobc:
Closed via https://github.com/spring-projects-experimental/spring-pulsar/commit/224eb0b3e2d9c2e1cb9466c2efda8f9dea851e24

PR #298, opened 2023-01-30
Pause/Resume further improvements

sobychacko:
Resovles https://github.com/spring-projects-experimental/spring-pulsar/issues/296

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainer.java
@@ -135,19 +135,15 @@ public class ConcurrentPulsarMessageListenerContainer<T> extends AbstractPulsarM
 	}
 
 	@Override
-	public void pause() {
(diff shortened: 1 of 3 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/listener/ConcurrentPulsarMessageListenerContainer.java
```diff
@@ -135,19 +135,15 @@ public List<DefaultPulsarMessageListenerContainer<T>> getContainers() {
 	}
 
 	@Override
-	public void pause() {
-		synchronized (this.lifecycleMonitor) {
-			super.pause();
-			this.containers.forEach(AbstractPulsarMessageListenerContainer::pause);
-		}
+	public void doPause() {
+		setPaused(true);
```
onobc, line 139: Should we only do this if it is. not currently paused already? Ditto below.

onobc (changes requested):
@sobychacko - this looks good -  one comment to address, but LGTM.

onobc (approved):
LGTM

Issue #297, opened 2023-01-30
Loop improvements while container paused


sobychacko:
One downside of this is that the loop will just continue spinning eventhough we know it will never get data while paused. To improve this we could instead leverage `Object.wait/notify` in order to let the code wait while releasing the thread resources.

_Originally posted by @onobc in https://github.com/spring-projects-experimental/spring-pulsar/pull/292#discussion_r1089828420_

PR #299, opened 2023-01-31
Container pause/resume improvements

sobychacko:
* When container is paused, ensure that we are not spinning CPU cycles by constantly trying to poll in a loop. When pausing the container, the polling waits until a condition is signaled by resuming.

Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/297

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
@@ -30,6 +30,9 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
@@ -90,6 +93,10 @@ public class DefaultPulsarMessageListenerContainer<T> extends AbstractPulsarMess
 	private final AtomicReference<Thread> listenerConsumerThread = new AtomicReference<>();
 
 	private final AtomicBoolean receiveInProgress = new AtomicBoolean();
+
+	private final Lock lockOnPause = new ReentrantLock();
+
+	private final Condition pausedCondition = this.lockOnPause.newCondition();
 
 	public DefaultPulsarMessageListenerContainer(PulsarConsumerFactory<? super T> pulsarConsumerFactory,
 			PulsarContainerProperties pulsarContainerProperties) {
@@ -207,6 +214,14 @@ public class DefaultPulsarMessageListenerContainer<T> extends AbstractPulsarMess
 			consumer.resume();
 		}
 		setPaused(false);
+		this.lockOnPause.lock();
+		try {
+			// signal the lock's condition to continue.
+			this.pausedCondition.signal();
+		}
+		finally {
+			this.lockOnPause.unlock();
+		}
 	}
 
 	private final class Listener implements SchedulingAwareRunnable {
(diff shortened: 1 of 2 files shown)


spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
```diff
@@ -87,6 +94,76 @@ void basicDefaultConsumer() throws Exception {
 		pulsarClient.close();
 	}
 
+	@Test
+	void containerPauseAndResumeFeatureUsingWaitAndNotify() throws Exception {
+		HashSet<String> topics = new HashSet<>();
```
onobc, line 99: This can be simplified w/ `Collections.singleton`.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
```diff
@@ -87,6 +94,76 @@ void basicDefaultConsumer() throws Exception {
...
+
+		container.pause();
+
+		Awaitility.await().until(container::isPaused);
+
+		container.resume();
+
+		Awaitility.await().until(() -> !container.isPaused());
+
+		assertThat(latchOnLockInvocation.await(10, TimeUnit.SECONDS)).isTrue();
```
onobc, line 158: This test verifies the use of the locks in the current impl really well. I wonder if we should also add an IT that sends/receives messages and make sure it behaves as expected?

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
```diff
@@ -87,6 +94,76 @@ void basicDefaultConsumer() throws Exception {
...
+		pulsarContainerProperties.setMessageListener((PulsarRecordMessageListener<?>) (consumer, msg) -> {
+		});
+		pulsarContainerProperties.setSchema(Schema.STRING);
+		DefaultPulsarMessageListenerContainer<String> container = new DefaultPulsarMessageListenerContainer<>(
+				pulsarConsumerFactory, pulsarContainerProperties);
+
+		Lock reentrantLock = new ReentrantLock();
+		Condition lockCondition = reentrantLock.newCondition();
+
+		Lock spyLock = spy(reentrantLock);
```
onobc, line 117: Let's keep an eye on any random failures from this test as I remember that we have had issues w/ the consumer tests that use the spy mocks.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -439,6 +454,21 @@ public void run() {
 						handleAcks(messages);
 					}
 				}
+				if (isPaused()) {
+					try {
+						// try acquiring the lock.
+						DefaultPulsarMessageListenerContainer.this.lockOnPause.lock();
```
onobc, line 460: Did you intend to have this w/in the try block?

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
```diff
@@ -439,6 +454,21 @@ public void run() {
 						handleAcks(messages);
 					}
 				}
+				if (isPaused()) {
```
onobc, line 457: Here is the current flow in this PR:
```
1) get messages
2) if no messages then continue back to top of loop
3) there are messages handle batch or single
4) now, check if we are paused
```
When the condition for 2) matches it will go back to 1) and skip the pause check altogether. 

If we move the pause check to the top of the loop I think it solves this. Either that, or also do a "handlePause" in the  `messages is null` case.

onobc (changes requested):
@sobychacko nice improvement and nice use of Lock/Condition for this. I added a couple comments.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: good catch.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainerTests.java
sobychacko: Ok, I will think about that (maybe outside of this PR).

spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
sobychacko: Very interesting. I didn't notice that flow where messages go null. I will update the PR.

onobc (approved):

PR #301, opened 2023-02-01
Basic Pulsar Reader API support in Spring Pulsar

sobychacko:

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarReaderFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarReaderFactory.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.core;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.pulsar.client.api.MessageId;
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.PulsarClientException;
+import org.apache.pulsar.client.api.Reader;
+import org.apache.pulsar.client.api.ReaderBuilder;
+import org.apache.pulsar.client.api.Schema;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * Default implementation of {@link PulsarReaderFactory}.
+ *
+ * @param <T> message type
+ * @author Soby Chacko
+ */
+public class DefaultPulsarReaderFactory<T> implements PulsarReaderFactory<T> {
+
+	private final PulsarClient pulsarClient;
+
+	private final Map<String, Object> readerConfig;
+
+	public DefaultPulsarReaderFactory(PulsarClient pulsarClient) {
+		this(pulsarClient, Collections.emptyMap());
+	}
+
+	public DefaultPulsarReaderFactory(PulsarClient pulsarClient, Map<String, Object> readerConfig) {
+		this.pulsarClient = pulsarClient;
+		this.readerConfig = readerConfig;
+	}
+
+	@Override
+	public Reader<T> createReader(@Nullable List<String> topics, @Nullable MessageId messageId, Schema<T> schema)
+			throws PulsarClientException {
+		ReaderBuilder<T> readerBuilder = this.pulsarClient.newReader(schema);
+		readerBuilder.topics(topics);
+		readerBuilder.startMessageId(messageId);
+
+		readerBuilder.loadConf(this.readerConfig);
+		return readerBuilder.create();
+	}
+
+}
--- a/spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultPulsarReaderFactoryTests.java
+++ b/spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultPulsarReaderFactoryTests.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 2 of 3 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarReaderFactory.java
```diff
@@ -0,0 +1,66 @@
...
+		this(pulsarClient, Collections.emptyMap());
+	}
+
+	public DefaultPulsarReaderFactory(PulsarClient pulsarClient, Map<String, Object> readerConfig) {
+		this.pulsarClient = pulsarClient;
+		this.readerConfig = readerConfig;
+	}
+
+	@Override
+	public Reader<T> createReader(@Nullable List<String> topics, @Nullable MessageId messageId, Schema<T> schema,
```
onobc, line 55: I have been going through the consumer factory API (like I did for the producer factory) and I have found that some of the APIs we provide are actually not usable as the consumer always requires topic and subscription info.

**Here is an example:**
For DPCF we have a constructor that takes in client + config map of default props. It has the following APIs to createConsumer:

* **API-1:** `createConsumer(Schema<T> schema)`
* **API-2:** `createConsumer(Schema<T> schema, @Nullable Collection<String> topics)`
* **API-3:** `createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
			@Nullable Map<String, String> properties, @Nullable List<ConsumerBuilderCustomizer<T>> customizers)`

**API-1** is only usable if there is a topic+sub in the config props (latter is unlikely)

**API-2** is only usable if there is a sub in the config props (unlikely)

**API-3** is the only one usable if topic+sub is not in the config props. The topic can be passed in via the `topics` param or via the customizer. The sub can be passed in via the customizer.

I think we should move to an API on the consumer/reader level that always includes the customizers. Currently, our main user of the factory - the listener containers - always pass in a customizer.

I think this will defintely be "less-is-more"  when/if users start using the factories direclty. Plus it is less surface area for us to test, maintain, explain.

For DPCF in the above example that would look like:

* `createConsumer(Schema<T> schema, @Nullable Collection<String> topics,
			@Nullable Map<String, String> properties,
      @Nullable ConsumerBuilderCustomizer<T>... customizers)`

**NOTE**: topics/properties are required to give users a chance to replace rather than append to the default topics/properties in the builder.

For the `PulsarReaderFactory` in this code proposal, that would change things to have only the following 2 methods:

* <init> `DefaultPulsarReaderFactory(PulsarClient pulsarClient, Map<String, Object> readerConfig`

* `Reader<T> createReader(Schema<T> schema, @Nullable ReaderBuilderCustomizer<T>... customizers)`

**NOTE**: The varargs allows them to easily pass 1 w/o having to wrap in list etc..

We will still be able to auto-configure the reader factory w/ this approach. This is exactly the approach that the Reactive consumer/reader take.

My end conclusion I have come to is that specifying individual properties on the APIs is brittle and can lead to confusing usage. Customizer approach will lead users to a more flexible, less-intrusive, and easier to use experience.

For example, one of the new tests uses the API like this currently:
```java
try (Reader<String> reader = readerFactory.createReader(
			List.of("basic-pulsar-reader-topic"), 
			MessageId.earliest, Schema.STRING, 
			Collections.emptyMap())) {
  ...
}
```

It would instead look like this:
```
try (var reader = readerFactory.createReader(Schema.STRING, (readerBuilder) -> {
  readerBuilder.topic("basic-pulsar-reader-topic");
  readerBuilder.startMessageId(MessageId.earliest);
});
```

spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultPulsarReaderFactoryTests.java
```diff
@@ -0,0 +1,103 @@
...
+		try (Reader<String> reader = pulsarReaderFactory.createReader(List.of("basic-pulsar-reader-topic"),
+				MessageId.earliest, Schema.STRING, Collections.emptyMap())) {
+
+			Map<String, Object> prodConfig = Map.of("topicName", "basic-pulsar-reader-topic");
+			PulsarProducerFactory<String> pulsarProducerFactory = new DefaultPulsarProducerFactory<>(pulsarClient,
+					prodConfig);
+			PulsarTemplate<String> pulsarTemplate = new PulsarTemplate<>(pulsarProducerFactory);
+			pulsarTemplate.send("hello john doe");
+
+			message = reader.readNext();
```
onobc, line 71: **[NIT]** This can be simplified by not declaring the method var on 61 and and just doing

```java
assertThat(reader.readNext())
    .extracting(Mesage::getValue).isEqualTo("hello john doe");
```

onobc (changes requested):
@sobychacko this looks good. I do have a request around the API based on what I have found w/ recent refactoring to the consumer factory.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarReaderFactory.java
sobychacko: In DPCF, we are only currently using the 3rd API (the other two are unused). I just started with a sketch of the reader factory API in this PR. We should definitely add the customizer option in a subsequent PR (soon, though). I actually like the approach of a few mandatory properties as part of the consumer/reader API (such as the schema, topic, and a general way to pass along the properties - similar to the Spring Kafka factory APIs). Then we should also provide the customizer approach as a final touchpoint for the user to customize anything - which should take priority if the user provides one. By choosing a few mandatory fields like the ones I mentioned and the starting message ID in the reader's case, we make our consumer/reader API's tight. No? Having only the customizer in the API feels a bit vague to me.

onobc: Btw @sobychacko , sorry for my initial giant manifesto comment 🤣 

> I actually like the approach of a few mandatory properties as part of the consumer/reader API (such as the schema, topic, and a general way to pass along the properties

In general I like that approach too. However, looking at where the consumer factory API has morphed, in that case it is not helpful as the user always requires a customizer unless they set subscription in the config props. That would move us back to adding yet another parameter in the API for subscription. That parameter is sometimes required (not needed if the sub is already in config props etc..) It is currently a very loose API. To make the "1-N primary properties where sometimes they are required API" work well we end up having to have a permutation of APIs where each API exposes only those primary properties (`schema, schema+topic, schema+topic+sub, schema+sub,schema+... etc.`) where the other option is to condense this down to a single API where each param is optional etc.. 

> Having only the customizer in the API feels a bit vague to me.
I thought the same at first. Then once I used it a couple of times it is the preferred way for me. It gives the user more power w/ the native pulsar builder interface that we end up ultimately trying to put a thin layer around w/ the convenience parameters. If you take a look at the Reactive counterpart (`ReactivePulsarReaderFactory`), it takes the customizer approach as well. 

That being said, the focus of my comments above are mainly the consumer factory. I am working on a PR to address that. I  do think this reader API is currently simpler than the consumer factory and this may work like it is. We will know soon enough.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarReaderFactory.java
```diff
@@ -0,0 +1,66 @@
...
+			@Nullable Map<String, Object> properties) throws PulsarClientException {
+		ReaderBuilder<T> readerBuilder = this.pulsarClient.newReader(schema);
+		readerBuilder.topics(topics);
+		readerBuilder.startMessageId(messageId);
+		Map<String, Object> config = new HashMap<>(this.readerConfig);
```
onobc, line 60: These (and therefore the default config props `readerConfig`) are not currently being used to build the reader. 

How do the properties passed into create() differ from the default readerConfig passed into `<init>`?

sobychacko: That is certainly a mistake - I will fix it tomorrow. Once again, I was just laying the foundations - this thing is not done by any means.

sobychacko: Actually, we don't need these properties (map) in the API, as we only need the instance field-level config map.

onobc: Cool. Yeh, I was thinking that we did not need to map either.

Issue #300, opened 2023-01-31
ReactivePulsarTemplate not ordering when schema not specified


onobc:
When the schema is not directly specified the RPT goes through `Flux.flatMapSequential`:
https://github.com/spring-projects-experimental/spring-pulsar/blob/ce96e7f5a994627375d5ae98244d9b312481bfb8/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java#L164

For some reason the resulting responses do not match the order of the messages in the input publisher.

This test shows the behavior:
https://github.com/spring-projects-experimental/spring-pulsar/blob/ce96e7f5a994627375d5ae98244d9b312481bfb8/spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplateTests.java#L129

onobc:
Fixed via https://github.com/spring-projects-experimental/spring-pulsar/commit/9fb8879491c2d96b2db6130cd33c926ae86f35e6

PR #303, opened 2023-02-01
Use switchOnFirst to determine the Schema and topic from the first message in doSendMany

cbornet:
In `doSendMany`, when we don't have an explicit Schema or topic, instead of using `sendOne` and create a sender for each message, we can use `switchOnFirst` to build the sender from the first message.
This has the following advantages:
* Can be used for both Schema and topic inference (currently only schema was supported with the sendOne workaround)
* Only one sender is created so the order is preserved even if there is no producer cache configured.
* We can remove `SendMessageBuilder::newMessages(Publisher<T>, Class<T> messageType)` as we get the messageType from the first message.


onobc (approved):
Nice simplification @cbornet. This also solves:
- https://github.com/spring-projects-experimental/spring-pulsar/issues/300
- and now sendMany gets the benefit of inferred schema always

PR #304, opened 2023-02-01
Make SendMessageBuilder interface more navigable

cbornet:
This way, if you use `newMessage`, you'll have `withMessageCustomizer ` and `send` returns a `Mono`.
Otherwise if you use `newMessages`, `send` returns a `Flux` (`sendMany` is renamed as `send` since there is no more name conflict)

### DIFF
--- a/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarOperations.java
+++ b/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarOperations.java
@@ -111,49 +111,43 @@ public interface ReactivePulsarOperations<T> {
 	Flux<MessageId> send(@Nullable String topic, Publisher<T> messages, @Nullable Schema<T> schema);
 
 	/**
-	 * Create a {@link SendMessageBuilder builder} for configuring and sending a message
-	 * reactively.
+	 * Create a {@link SendOneMessageBuilder builder} for configuring and sending a
+	 * message reactively.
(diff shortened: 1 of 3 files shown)


spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarOperations.java
```diff
@@ -111,59 +111,53 @@
...
 
 	/**
 	 * Builder that can be used to configure and send a message. Provides more options
 	 * than the send methods provided by {@link ReactivePulsarOperations}.
 	 *
+	 * @param <O> the builder type
 	 * @param <T> the message payload type
 	 */
-	interface SendMessageBuilder<T> {
+	sealed interface SendMessageBuilder<O, T> permits SendOneMessageBuilder, SendManyMessageBuilder {
```
onobc, line 146: Nice use of `sealed/permits`.

onobc (approved):
@cbornet this definitely helps w/ the navigability/usability of the builder. When I added "messageType"  to "newMessages" I quickly realized it "split" the API when I was using/documenting it. When you find yourself saying "Use these method IF you are sending 1 BUT use these methods IF you are sending >1" you know it needs refactoring :) 

I had planned on circling back and doing something like this in the next 1-2 weeks. Glad you beat me to it though. Well done. 

Do you mind doing a rebase and deal w/ conflict on previous PR merge? I am on limited power & internet for the next 24-48 hrs and my availability is unpredictable.

PR #306, opened 2023-02-02
Tests for shared subscriptions

sobychacko:
Resolves https://github.com/spring-projects-experimental/spring-pulsar/issues/305

### DIFF
--- a/spring-pulsar/src/test/java/org/springframework/pulsar/core/SharedSubscriptionConsumerTests.java
+++ b/spring-pulsar/src/test/java/org/springframework/pulsar/core/SharedSubscriptionConsumerTests.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


spring-pulsar/src/test/java/org/springframework/pulsar/core/SharedSubscriptionConsumerTests.java
```diff
@@ -0,0 +1,169 @@
...
+
+		pulsarTemplate.newMessage("hello john doe").sendAsync();
+		pulsarTemplate.newMessage("hello alice doe").sendAsync();
+		pulsarTemplate.newMessage("hello buzz doe").sendAsync();
+
+		Awaitility.await().atMost(Duration.ofSeconds(30)).until(() -> latch1.await(10, TimeUnit.SECONDS));
+		Awaitility.await().atMost(Duration.ofSeconds(30)).until(() -> latch2.await(10, TimeUnit.SECONDS));
+		Awaitility.await().atMost(Duration.ofSeconds(30)).until(() -> latch3.await(10, TimeUnit.SECONDS));
+
+		container1.stop();
```
onobc, line 92: I know in the other listener tests we need to go back and put the resources in try-with-resources. I think it would be a good idea to do this now in this new test while there is only 3 tests in here. Wdyt?

spring-pulsar/src/test/java/org/springframework/pulsar/core/SharedSubscriptionConsumerTests.java
```diff
@@ -0,0 +1,169 @@
...
+		Map<String, Object> prodConfig = Map.of("topicName", "shared-subscription-single-msg-test-topic");
+		DefaultPulsarProducerFactory<String> pulsarProducerFactory = new DefaultPulsarProducerFactory<>(pulsarClient,
+				prodConfig);
+		PulsarTemplate<String> pulsarTemplate = new PulsarTemplate<>(pulsarProducerFactory);
+
+		pulsarTemplate.newMessage("hello john doe").sendAsync();
+		pulsarTemplate.newMessage("hello alice doe").sendAsync();
+		pulsarTemplate.newMessage("hello buzz doe").sendAsync();
+
+		Awaitility.await().atMost(Duration.ofSeconds(30)).until(() -> latch1.await(10, TimeUnit.SECONDS));
```
onobc, line 88: I think we should line up the timeouts as in a failure case it will always wait 30s for each of these, rather than 10. I wonder if there is another way we can use Awaitility so that these awaits are not done in serial?

onobc (changes requested):
Nice test addition @sobychacko . 

A couple of comments to address, but looks great.

PR #307, opened 2023-02-02
Bump the Pulsar reactive client to 0.2.0

cbornet:

### DIFF
--- a/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
+++ b/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 
 import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.Schema;
+import org.apache.pulsar.reactive.client.api.MessageSendResult;
 import org.apache.pulsar.reactive.client.api.MessageSpec;
 import org.apache.pulsar.reactive.client.api.MessageSpecBuilder;
 import org.apache.pulsar.reactive.client.api.ReactiveMessageSender;
@@ -62,9 +63,10 @@ public class ReactivePulsarTemplate<T> implements ReactivePulsarOperations<T> {
 	}
 
 	/**
-	 * Construct a template instance with a custom schema resolver.
+	 * Construct a template instance with a custom schema resolver and a custom topic
+	 * resolver.
 	 * @param reactiveMessageSenderFactory the factory used to create the backing Pulsar
-	 * @param schemaResolver the schema resolver to use reactive senders
+	 * @param schemaResolver the schema resolver to use
 	 * @param topicResolver the topic resolver to use
 	 */
 	public ReactivePulsarTemplate(ReactivePulsarSenderFactory<T> reactiveMessageSenderFactory,
@@ -95,22 +97,23 @@ public class ReactivePulsarTemplate<T> implements ReactivePulsarOperations<T> {
 	}
 
 	@Override
-	public Flux<MessageId> send(Publisher<T> messages) {
+	public Flux<MessageSendResult<T>> send(Publisher<MessageSpec<T>> messages) {
 		return send(null, messages);
 	}
 
 	@Override
-	public Flux<MessageId> send(Publisher<T> messages, @Nullable Schema<T> schema) {
+	public Flux<MessageSendResult<T>> send(Publisher<MessageSpec<T>> messages, @Nullable Schema<T> schema) {
 		return doSendMany(null, Flux.from(messages), schema, null);
 	}
 
 	@Override
-	public Flux<MessageId> send(@Nullable String topic, Publisher<T> messages) {
+	public Flux<MessageSendResult<T>> send(@Nullable String topic, Publisher<MessageSpec<T>> messages) {
 		return doSendMany(topic, Flux.from(messages), null, null);
 	}
 
 	@Override
-	public Flux<MessageId> send(@Nullable String topic, Publisher<T> messages, @Nullable Schema<T> schema) {
+	public Flux<MessageSendResult<T>> send(@Nullable String topic, Publisher<MessageSpec<T>> messages,
+			@Nullable Schema<T> schema) {
 		return doSendMany(topic, Flux.from(messages), schema, null);
 	}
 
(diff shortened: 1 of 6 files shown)


spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
```diff
@@ -137,13 +140,14 @@ private Mono<MessageId> doSend(@Nullable String topic, T message, @Nullable Sche
...
+	private Flux<MessageSendResult<T>> doSendMany(@Nullable String topic, Flux<MessageSpec<T>> messages,
+			@Nullable Schema<T> schema) {
 		return messages.switchOnFirst((firstSignal, messageFlux) -> {
-			T firstMessage = firstSignal.get();
+			MessageSpec<T> firstMessage = firstSignal.get();
 			if (firstMessage != null && firstSignal.isOnNext()) {
-				String topicName = resolveTopic(topic, firstMessage.getClass());
-				ReactiveMessageSender<T> sender = createMessageSender(topicName, firstMessage, schema, null);
-				return messageFlux.map(MessageSpec::of).as(sender::sendMany)
+				String topicName = resolveTopic(topic, firstMessage.getValue().getClass());
```
cbornet, line 148: The good news is that with `MessageSpec` we can send tombstone/null values (it's not possible with a Flux<T> as a Flux can't emit null.
The bad news is that if the first message is a tombstone, we can't determine the topic or the schema (and this code will NPE 😬

build.gradle
```diff
@@ -24,6 +24,12 @@ allprojects {
 	configurations.all {
 		resolutionStrategy.cacheChangingModulesFor 0, "minutes"
 	}
+
+	repositories {
```
onobc, line 28: This is a temporary override of the regular snapshot repo (defined [here](https://github.com/spring-projects-experimental/spring-pulsar/blob/b9710b947a258925e80a42393a0976a07a4327a7/buildSrc/src/main/groovy/io/spring/gradle/convention/RepositoryConventionPlugin.groovy#L52)) and will be removed once `0.2.0` is released, correct? This allows us to move forward and get a green CI run now.

spring-pulsar-docs/src/main/asciidoc/reactive-quick-tour.adoc
```diff
@@ -45,7 +45,7 @@ public class ReactiveSpringPulsarHelloWorld {
 
     @Bean
     ApplicationRunner runner(ReactivePulsarTemplate<String> pulsarTemplate) {
-        return (args) -> pulsarTemplate.send("hello-pulsar-topic", "Hello Reactive Pulsar World!").subscribe();
+        return (args) -> pulsarTemplate.send("hello-pulsar-topic", Flux.just(MessageSpec.of("Hello Reactive Pulsar World!"))).subscribe();
```
onobc, line 48: I think the previous sample use w/ single message may be less to digest on the quick-tour.

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
```diff
@@ -62,9 +63,10 @@ public ReactivePulsarTemplate(ReactivePulsarSenderFactory<T> reactiveMessageSend
 	}
 
 	/**
-	 * Construct a template instance with a custom schema resolver.
+	 * Construct a template instance with a custom schema resolver and a custom topic
```
onobc, line 66: Good catch - thx for the update.

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
onobc: I like the good news better! 🤣 That is great news w/ the tombstone ability. 

I feel like we should add a null check on the value and log something accordingly? 

So basically we support tombstone now w/ send many w/ the one caveat that they can not be the 1st message in the published stream. Can we "switchOnSecond" in that case? I don't know if that is even a thing - just thinking of ideas Even if that worked, could they have 2 null values back to back at position 1 and 2?

onobc (changes requested):
@cbornet good call on pre-loading this PR w/ the changes to the builder - it made this much easier to digest. 

LGTM! I added a couple of comments. After we resolve those then this is good to go.

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
cbornet: > Even if that worked, could they have 2 null values back to back at position 1 and 2?

They could have an infinite stream of nulls 😉

build.gradle
cbornet: Exactly. That's why the PR is in Draft mode atm

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarTemplate.java
cbornet: Or a Flux that emits just one null MS and completes.

cbornet: BTW we have the same problem for the `send(T message)` operation.

onobc: True and true. 
This PR is already above/beyond its initial goal. I have created https://github.com/spring-projects-experimental/spring-pulsar/issues/311 to follow up w/ the docs for sending tombstones.

build.gradle
```diff
@@ -24,6 +24,12 @@ allprojects {
...
+		maven {
```
onobc, line 29: Looks like we can pull this now @cbornet 

https://repo.maven.apache.org/maven2/org/apache/pulsar/pulsar-client-reactive-api/0.2.0/

onobc (changes requested):
https://github.com/spring-projects-experimental/spring-pulsar/pull/307/files#r1098152729

build.gradle
cbornet: Indeed. Done.

onobc (approved):
@cbornet - lets go!

PR #320, opened 2023-02-06
Trim ReactivePulsarSenderFactory API

onobc:
### What
Removes the `createProducer/Sender(Schema)` as it is likely not that useful as most users will specify a topic and not rely on the default topic. 

### Why
The thinking is that the `Reactive/PulsarTemplate` will be the users primary sending API. The producer/sender API is therefore secondary in that sense. This allows us to limit the producer/sender API surface area and remove any questionable APIs in `1.0.0` and let users drive the shape of the API `post-1.0.0`.

Additionally:
* Consistency: A goal is to keep the imperative and reactive producer/sender consistent (as much as possible). I already did this same change to the imperative PPF.  This PR applies the same treatment to the Reactive counterpart.

cc: @cbornet if you agree (don't have an issue w/ this approach) I am planning on removing the schema-only API in the consumer/readers as well:

https://github.com/spring-projects-experimental/spring-pulsar/blob/f211a3490ba40049036f7b78ba63660f984aeabe/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarConsumerFactory.java#L37

RPRF and RPCF (https://github.com/spring-projects-experimental/spring-pulsar/blob/f211a3490ba40049036f7b78ba63660f984aeabe/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/ReactivePulsarReaderFactory.java#L37)

### DIFF
--- a/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/DefaultReactivePulsarSenderFactory.java
+++ b/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/DefaultReactivePulsarSenderFactory.java
@@ -18,6 +18,7 @@ package org.springframework.pulsar.reactive.core;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import org.apache.pulsar.client.api.PulsarClient;
 import org.apache.pulsar.client.api.Schema;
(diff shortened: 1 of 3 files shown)


cbornet:
👌

PR #324, opened 2023-02-06
Add support for tombstone/null values

### LINKED ISSUE #311: Reactive client ability to send tombstone
The ability to send tombstone (`null`) messages in `ReactivePulsarTemplate.send(Flux)` was enabled via https://github.com/spring-projects-experimental/spring-pulsar/pull/307. However, there is a caveat in that if the 1st record in the `Flux` is the tombstone record it will NPE during topic resolution.  The same problem already existed for the singular `ReactivePulsarTemplate.send(null)`. 

The limitation is that if you want to send tombstones you must specify the topic on the send operation or via the config props (default topic). 

The work here is to:
- [ ] document this feature and its limitations (add a ~`Sending Tombstones` section in docs)
- [ ] ensure that the `RPT.send(T)` w/ topic specified does not NPE on tombstone record  (add test(s))
- [ ] ensure that the `RPT.sendMany(Publisher<T>)` w/ topic specified does not NPE when tombstone record is 1st in flux (add test(s))

cbornet:
Fix #311

### DIFF
--- a/spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinderResolveSchemaTests.java
+++ b/spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinderResolveSchemaTests.java
@@ -17,9 +17,13 @@
 package org.springframework.pulsar.spring.cloud.stream.binder;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatExceptionOfType;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 
+import org.apache.pulsar.client.api.Schema;
 import org.apache.pulsar.common.schema.KeyValue;
 import org.apache.pulsar.common.schema.SchemaType;
 import org.junit.jupiter.api.Nested;
@@ -31,6 +35,7 @@ import org.junit.jupiter.params.provider.EnumSource.Mode;
 import org.springframework.core.ResolvableType;
 import org.springframework.pulsar.core.PulsarConsumerFactory;
 import org.springframework.pulsar.core.PulsarTemplate;
+import org.springframework.pulsar.core.Resolved;
 import org.springframework.pulsar.core.SchemaResolver;
 import org.springframework.pulsar.spring.cloud.stream.binder.provisioning.PulsarTopicProvisioner;
 
@@ -50,21 +55,24 @@ public class PulsarMessageChannelBinderResolveSchemaTests {
 	@ParameterizedTest
 	@EnumSource(mode = Mode.MATCH_NONE, names = "^(AUTO.*|AVRO|JSON|KEY_VALUE|NONE|PROTOBUF.*)$")
 	void primitiveSchemaTypes(SchemaType schemaType) {
+		doReturn(Resolved.of(Schema.STRING)).when(resolver).resolveSchema(schemaType, null);
 		binder.resolveSchema(schemaType, null, null, null);
-		verify(resolver).getSchema(schemaType, null);
+		verify(resolver).resolveSchema(schemaType, null);
 	}
 
 	@ParameterizedTest
 	@EnumSource(mode = Mode.MATCH_ALL, names = "^(JSON|AVRO|PROTOBUF)$")
 	void structSchemaTypes(SchemaType schemaType) {
+		doReturn(Resolved.of(Schema.STRING)).when(resolver).resolveSchema(eq(schemaType), any());
 		binder.resolveSchema(schemaType, Foo.class, null, null);
-		verify(resolver).getSchema(schemaType, ResolvableType.forClass(Foo.class));
+		verify(resolver).resolveSchema(schemaType, ResolvableType.forClass(Foo.class));
 	}
 
 	@Test
 	void keyValueSchemaType() {
+		doReturn(Resolved.of(Schema.STRING)).when(resolver).resolveSchema(eq(SchemaType.KEY_VALUE), any());
 		binder.resolveSchema(SchemaType.KEY_VALUE, null, Foo.class, Bar.class);
-		verify(resolver).getSchema(SchemaType.KEY_VALUE,
+		verify(resolver).resolveSchema(SchemaType.KEY_VALUE,
 				ResolvableType.forClassWithGenerics(KeyValue.class, Foo.class, Bar.class));
 	}
 
@@ -94,21 +102,24 @@ public class PulsarMessageChannelBinderResolveSchemaTests {
 
 		@Test
 		void withMesssageType() {
+			doReturn(Resolved.of(Schema.STRING)).when(resolver).resolveSchema(eq(SchemaType.NONE), any());
 			binder.resolveSchema(SchemaType.NONE, Foo.class, null, null);
-			verify(resolver).getSchema(SchemaType.NONE, ResolvableType.forClass(Foo.class));
+			verify(resolver).resolveSchema(SchemaType.NONE, ResolvableType.forClass(Foo.class));
 		}
 
 		@Test
 		void withKeyAndValueTypes() {
+			doReturn(Resolved.of(Schema.STRING)).when(resolver).resolveSchema(eq(SchemaType.NONE), any());
 			binder.resolveSchema(SchemaType.NONE, null, Foo.class, Bar.class);
-			verify(resolver).getSchema(SchemaType.NONE,
+			verify(resolver).resolveSchema(SchemaType.NONE,
 					ResolvableType.forClassWithGenerics(KeyValue.class, Foo.class, Bar.class));
 		}
 
 		@Test
 		void withMessageTypeAndKeyAndValueTypes() {
+			doReturn(Resolved.of(Schema.STRING)).when(resolver).resolveSchema(eq(SchemaType.NONE), any());
 			binder.resolveSchema(SchemaType.NONE, Foo.class, String.class, Bar.class);
-			verify(resolver).getSchema(SchemaType.NONE, ResolvableType.forClass(Foo.class));
+			verify(resolver).resolveSchema(SchemaType.NONE, ResolvableType.forClass(Foo.class));
 		}
 
 		@Test
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarProducerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarProducerFactory.java
@@ -126,8 +126,7 @@ public class DefaultPulsarProducerFactory<T> implements PulsarProducerFactory<T>
 
 	protected String resolveTopicName(String userSpecifiedTopic) {
 		String defaultTopic = Objects.toString(getProducerConfig().get("topicName"), null);
-		return this.topicResolver.resolveTopic(userSpecifiedTopic, () -> defaultTopic).orElseThrow(
-				() -> new IllegalArgumentException("Topic must be specified when no default topic is configured"));
+		return this.topicResolver.resolveTopic(userSpecifiedTopic, () -> defaultTopic).orElseThrow();
 	}
 
 	@Override
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultTopicResolver.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultTopicResolver.java
@@ -71,17 +71,20 @@ public class DefaultTopicResolver implements TopicResolver {
 
 	@Override
 	public Optional<String> resolveTopic(@Nullable String userSpecifiedTopic, Supplier<String> defaultTopicSupplier) {
-		return doResolveTopic(userSpecifiedTopic, null, defaultTopicSupplier);
+		if (StringUtils.hasText(userSpecifiedTopic)) {
+			return Optional.of(userSpecifiedTopic);
+		}
+		return Optional.ofNullable(defaultTopicSupplier.get());
 	}
 
 	@Override
-	public <T> Optional<String> resolveTopic(@Nullable String userSpecifiedTopic, T message,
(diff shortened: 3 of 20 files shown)


spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultTopicResolverTests.java
```diff
@@ -110,7 +110,7 @@ static Stream<Arguments> resolveByMessageTypeProvider() {
 				arguments("complexMessageWithUserTopic", userTopic, Foo.class, defaultTopic, userTopic),
 				arguments("complexMessageNoUserTopic", null, Foo.class, defaultTopic, fooTopic),
 				arguments("nullMessageWithUserTopicAndDefault", userTopic, null, defaultTopic, userTopic),
-				arguments("nullMessageWithDefault", null, null, defaultTopic, defaultTopic),
+				arguments("nullMessageWithDefault", null, null, defaultTopic, null),
```
cbornet, line 113: Is it OK ?
Or do we prefer to return the default topic when the message is null ?

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultTopicResolver.java
```diff
@@ -91,6 +94,9 @@ private Optional<String> doResolveTopic(@Nullable String userSpecifiedTopic, @Nu
 		if (StringUtils.hasText(userSpecifiedTopic)) {
 			return Optional.of(userSpecifiedTopic);
 		}
+		if (messageType == null) {
```
onobc, line 97: This change will break [this about to merged change](https://github.com/spring-projects-experimental/spring-pulsar/pull/322/files#diff-60aaf1bc942ff45a5fa56d82ad06cfe23c20b4ffa9d1dd9f2d23a2818b969677R98). 

I switched to use the TR in the sender/producer so that all of the flows go through the resolver. If the user changed the TR impl then it would be nice if all flows go through there. The sender/producer do not have the message on their own at this point so the default impl just does `userTopic OR defaultTopic` resolution. In this proposed changed that would instead do `userTopic or empty`.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultTopicResolver.java
```diff
@@ -71,17 +71,20 @@ public Map<Class<?>, String> getCustomTopicMappings() {
 
 	@Override
 	public Optional<String> resolveTopic(@Nullable String userSpecifiedTopic, Supplier<String> defaultTopicSupplier) {
-		return doResolveTopic(userSpecifiedTopic, null, defaultTopicSupplier);
+		if (StringUtils.hasText(userSpecifiedTopic)) {
```
onobc, line 74: Nice catch here as this would otherwise have broken my recent change to the sender/producer to call through the TR as well.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -177,13 +180,22 @@ private MessageId doSend(@Nullable String topic, T message, @Nullable Schema<T>
...
-			@Nullable Collection<String> encryptionKeys,
+	private CompletableFuture<MessageId> doSendAsync(@Nullable String topic, @Nullable T message,
+			@Nullable Schema<T> schema, @Nullable Collection<String> encryptionKeys,
 			@Nullable TypedMessageBuilderCustomizer<T> typedMessageBuilderCustomizer,
 			@Nullable ProducerBuilderCustomizer<T> producerCustomizer) throws PulsarClientException {
 		String defaultTopic = Objects.toString(this.producerFactory.getProducerConfig().get("topicName"), null);
-		String topicName = this.topicResolver.resolveTopic(topic, message, () -> defaultTopic).orElseThrow(
-				() -> new IllegalArgumentException("Topic must be specified when no default topic is configured"));
+		String topicName = this.topicResolver.resolveTopic(topic, message, () -> defaultTopic).orElseThrow(() -> {
+			if (this.topicResolver.getClass().equals(DefaultTopicResolver.class)) {
```
onobc, line 189: Yeh, I saw this before, it made me think of returning an `Either` (result/error) **OR** adding a `forceResolveTopic` (or force param or something like that) to signal to throw exception if not resolved. As the `if message type .. use this ex else use this ex" is definitely an impl. concern of the DTR.  

On one hand I am ok w/ just keeping this simple and using a generic msg altogether.
On the other hand, the "when null" could be helpful as that is the tombstone case and it may give users clue as to what the issue is.

I'm fine w/ whatever approach you end up going w/. If you are good w/ this one - simply resolve the comment. Thx

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -192,13 +204,19 @@ private CompletableFuture<MessageId> doSendAsync(@Nullable String topic, T messa
...
+			try {
+				messageBuilder = producer.newMessage().value(message);
+				if (typedMessageBuilderCustomizer != null) {
+					typedMessageBuilderCustomizer.customize(messageBuilder);
+				}
+				// propagate props to message
+				senderContext.properties().forEach(messageBuilder::property);
+			}
+			catch (Exception e) {
+				ProducerUtils.closeProducerAsync(producer, this.logger);
```
onobc, line 217: Oh, what brought this up?

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
```diff
@@ -227,12 +245,15 @@ private Observation newObservation(PulsarMessageSenderContext senderContext) {
...
-	private Producer<T> prepareProducerForSend(@Nullable String topic, T message, @Nullable Schema<T> schema,
+	private Producer<T> prepareProducerForSend(@Nullable String topic, @Nullable T message, @Nullable Schema<T> schema,
 			@Nullable Collection<String> encryptionKeys, @Nullable ProducerBuilderCustomizer<T> producerCustomizer)
 			throws PulsarClientException {
-		if (schema == null) {
-			schema = Objects.requireNonNull(this.schemaResolver.getSchema(message),
-					"Schema must not be null - expecting at least a default schema");
+		Schema<T> resolvedSchema = schema == null ? this.schemaResolver.getSchema(message) : schema;
+		if (resolvedSchema == null) {
+			if (this.schemaResolver.getClass().equals(DefaultSchemaResolver.class)) {
```
onobc, line 253: Ah yeh, this again. 😆 

Should we add a `SchemaResolver.exceptionForUnresolvableMesage(message)` and `TopicResolver.exceptionForUnresolvable(message)` ? Or something like that?

spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultTopicResolverTests.java
onobc: I'm not sure. Up until now we have returned default when message is null. 

1) If we return the default topic then tombstones will go to the default topic (which is probably not the best choice). 
2) If we return empty then they get a clear answer that they need to specify the topic.

Of these I would pick 2) as bad things can happen if we start shipping nulls to a topic a user configured as the "default" (but did not mean for the null records).

Where else can the message be null that calls in from here? What about the consumer containers? Will this affect them in any way? They have message type though IIRC.

onobc (changes requested):
@cbornet as usual, looks great!

We have a few comments to wrap up discussion on so marking as RC until we close those.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: **OR** we could change the TR signature to return ~ `ResolveTopicResult` which could contain an optional resolved topic OR a reason why it could not be resolved. That would remove the need for special message casing here and allow custom impls ability to tell users about why their custom resolver did not work. Wdyt?

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
onobc: **OR** we could change the TR signature to return ~ `ResolveTopicResult` which could contain an optional resolved topic OR a reason why it could not be resolved. That would remove the need for special message casing here and allow custom impls ability to tell users about why their custom resolver did not work. Wdyt?

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
cbornet: If the message type doesn't match the Producer one (eg. the message type is Foo.class and the producer is Schema.BYTES), `producer.newMessage().value(message)` throws a cast exception and the producer is not closed.

spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultTopicResolverTests.java
cbornet: On the consumer path, we get the type by reflection so it won't be null.

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
cbornet: Good idea. I've added a `Resolved<T>` wrapper for this. PTAL

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java
cbornet: I also used `Resolved<T>` for the schema

spring-pulsar/src/main/java/org/springframework/pulsar/core/Resolved.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package org.springframework.pulsar.core;
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import org.springframework.lang.Nullable;
+
+public final class Resolved<T> {
```
onobc, line 24: Nice re-use w/ `<T>`.

onobc, line 24: Do you mind adding javadoc w/ `@author`?

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultTopicResolver.java
```diff
@@ -70,28 +69,40 @@ public Map<Class<?>, String> getCustomTopicMappings() {
...
 
 	@Override
-	public Optional<String> resolveTopic(@Nullable String userSpecifiedTopic, Class<?> messageType,
+	public Resolved<String> resolveTopic(@Nullable String userSpecifiedTopic, @Nullable Class<?> messageType,
 			Supplier<String> defaultTopicSupplier) {
 		return doResolveTopic(userSpecifiedTopic, messageType, defaultTopicSupplier);
 	}
 
-	private Optional<String> doResolveTopic(@Nullable String userSpecifiedTopic, @Nullable Class<?> messageType,
+	private Resolved<String> doResolveTopic(@Nullable String userSpecifiedTopic, @Nullable Class<?> messageType,
```
onobc, line 95: Should we open visiblity to protected here like we did for SR?

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarProducerFactory.java
```diff
@@ -126,8 +126,7 @@ protected Producer<T> doCreateProducer(Schema<T> schema, @Nullable String topic,
 
 	protected String resolveTopicName(String userSpecifiedTopic) {
 		String defaultTopic = Objects.toString(getProducerConfig().get("topicName"), null);
-		return this.topicResolver.resolveTopic(userSpecifiedTopic, () -> defaultTopic).orElseThrow(
-				() -> new IllegalArgumentException("Topic must be specified when no default topic is configured"));
+		return this.topicResolver.resolveTopic(userSpecifiedTopic, () -> defaultTopic).orElseThrow();
```
onobc, line 129: Nice!

spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinderResolveSchemaTests.java
```diff
@@ -41,30 +46,33 @@
  */
 public class PulsarMessageChannelBinderResolveSchemaTests {
 
-	private SchemaResolver resolver = mock(SchemaResolver.class);
+	private final SchemaResolver resolver = mock(SchemaResolver.class);
```
onobc, line 49: **[NIT]** Don't worry about removing now @cbornet - but we have avoided littering the code w/ finals. I know IDEA has warned that we could make this final. We landed on a guideline of making instance vars for immutable classes final and constants as well. Its mostly about consistency - IMO either we do it everywhere or not at all or in a set of prescribed cases that add value.

spring-pulsar/src/main/java/org/springframework/pulsar/core/Resolved.java
cbornet: Sure

onobc (changes requested):
@cbornet I really like the `Resolved` concept - turned out to fit the bill well. 

I added a couple of minor comments - then I promise I will stop bikeshedding ;P

spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinderResolveSchemaTests.java
cbornet: done

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultTopicResolver.java
cbornet: done

onobc (approved):

onobc:
Nice work @cbornet ! Thank you.

PR #325, opened 2023-02-06
Use Schema.JSON instead of Schema.BYTES as default schema

### LINKED ISSUE #323: SchemaResolver default to Schema.JSON
For discussion.
Currently, if the Schema is not specified and there is no custom schema mapping, the schema resolver defaults to BYTES.
It could be interesting to default to JSON instead using Schema.JSON(messageType).

cbornet:
Fix #323

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
@@ -130,7 +130,11 @@ public class DefaultSchemaResolver implements SchemaResolver {
 
 	@Nullable
 	private Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
-		return this.customSchemaMappings.getOrDefault(messageClass, (returnDefault ? Schema.BYTES : null));
+		Schema<?> schema = this.customSchemaMappings.get(messageClass);
+		if (schema == null && returnDefault) {
+			return messageClass != null ? Schema.JSON(messageClass) : Schema.BYTES;
+		}
+		return schema;
 	}
 
 	@Override
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
@@ -39,6 +39,7 @@ import org.apache.pulsar.common.schema.KeyValueEncodingType;
 import org.apache.pulsar.common.schema.SchemaType;
 
 import org.springframework.core.ResolvableType;
+import org.springframework.core.log.LogAccessor;
 import org.springframework.lang.Nullable;
 
 import com.google.protobuf.GeneratedMessageV3;
@@ -55,6 +56,8 @@ import com.google.protobuf.GeneratedMessageV3;
  * @author Chris Bono
  */
 public class DefaultSchemaResolver implements SchemaResolver {
+
+	private final LogAccessor logger = new LogAccessor(this.getClass());
 
 	private static final Map<Class<?>, Schema<?>> BASE_SCHEMA_MAPPINGS = new HashMap<>();
 	static {
@@ -129,10 +132,18 @@ public class DefaultSchemaResolver implements SchemaResolver {
 	}
 
 	@Nullable
-	private Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
+	protected Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
 		Schema<?> schema = this.customSchemaMappings.get(messageClass);
 		if (schema == null && returnDefault) {
-			return messageClass != null ? Schema.JSON(messageClass) : Schema.BYTES;
+			if (messageClass != null) {
+				try {
+					return Schema.JSON(messageClass);
+				}
+				catch (Exception e) {
+					this.logger.debug(e, "Failed to create JSON schema for " + messageClass.getName());
+				}
+			}
+			return Schema.BYTES;
 		}
 		return schema;
 	}
--- a/spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultSchemaResolverTests.java
+++ b/spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultSchemaResolverTests.java
@@ -39,7 +39,6 @@ import org.apache.pulsar.client.impl.schema.ProtobufSchema;
 import org.apache.pulsar.common.schema.KeyValue;
 import org.apache.pulsar.common.schema.KeyValueEncodingType;
 import org.apache.pulsar.common.schema.SchemaType;
-import org.apache.pulsar.shade.org.apache.avro.AvroTypeException;
 import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
@@ -140,10 +139,8 @@ class DefaultSchemaResolverTests {
 			resolver.addCustomSchemaMapping(Bar.class, Schema.STRING);
 			assertThat(resolver.getSchema(new Foo("foo1"))).isSameAs(fooSchema);
 			assertThat(resolver.getSchema(new Bar<>("bar1"))).isEqualTo(Schema.STRING);
-			assertThat(resolver.getSchema(new Zaa("zaa1")).getSchemaInfo()).satisfies(schemaInfo -> {
-				assertThat(schemaInfo.getType()).isEqualTo(SchemaType.JSON);
-				assertThat(schemaInfo.getSchema()).isEqualTo(Schema.JSON(Zaa.class).getSchemaInfo().getSchema());
-			});
+			assertThat(resolver.getSchema(new Zaa("zaa1")).getSchemaInfo())
+					.isEqualTo(Schema.JSON(Zaa.class).getSchemaInfo());
 		}
 
 	}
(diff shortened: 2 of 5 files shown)


spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
```diff
@@ -336,7 +355,53 @@ SendTestArgs producerCustomizer() {
 
 	}
 
-	record Foo(String foo, String bar) {
+	public static class Foo {
```
cbornet, line 358: AVRO doesn't work with records atm

onobc: Yeh, it is unfortunate. You will see a `User` and `User2` scattered about in a few tests because of this.

onobc: Doing this 2x now has me on the verge of bringing in Lombok just for testing and `@Data` in this case. The next time I find myself doing this, its Lombok time. :)

spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultSchemaResolverTests.java
```diff
@@ -305,13 +310,15 @@ void primitiveKeyValueMessageType() {
...
-			void customKeyValueMessageTypeDefaultsToBytesSchema() {
-				ResolvableType kvType = ResolvableType.forClassWithGenerics(KeyValue.class, Foo.class, Bar.class);
+			void customKeyValueMessageTypeDefaultsToJSONSchema() {
+				ResolvableType kvType = ResolvableType.forClassWithGenerics(KeyValue.class, Foo.class, Zaa.class);
 				assertThat(resolver.getSchema(SchemaType.NONE, kvType))
 						.asInstanceOf(InstanceOfAssertFactories.type(KeyValueSchema.class))
 						.satisfies((keyValueSchema -> {
-							assertThat(keyValueSchema.getKeySchema()).isEqualTo(Schema.BYTES);
-							assertThat(keyValueSchema.getValueSchema()).isEqualTo(Schema.BYTES);
+							assertThat(keyValueSchema.getKeySchema().getSchemaInfo())
```
onobc, line 318: So this change in defaults REALLY helps out in the key-value case as this prevents having to specify the key-value and type-value mappings.

spring-pulsar/src/test/java/org/springframework/pulsar/core/DefaultSchemaResolverTests.java
```diff
@@ -189,11 +193,12 @@ static Stream<Arguments> primitiveMessageTypesProvider() {
...
 		void customMessageTypes() {
 			assertThat(resolver.getSchema(Foo.class, false)).isNull();
-			assertThat(resolver.getSchema(Foo.class, true)).isEqualTo(Schema.BYTES);
+			assertThat(resolver.getSchema(Foo.class, true).getSchemaInfo())
+					.isEqualTo(Schema.JSON(Foo.class).getSchemaInfo());
 			resolver.addCustomSchemaMapping(Foo.class, Schema.STRING);
 			assertThat(resolver.getSchema(Foo.class, false)).isEqualTo(Schema.STRING);
 			assertThat(resolver.getSchema(Bar.class, false)).isNull();
-			assertThat(resolver.getSchema(Bar.class, true)).isEqualTo(Schema.BYTES);
+			assertThatExceptionOfType(AvroTypeException.class).isThrownBy(() -> resolver.getSchema(Bar.class, true));
```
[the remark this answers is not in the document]
onobc: But it's fine for `Bar`. Hmmm? I wonder how limiting this will be in practice? Many pojos are probably typed. 

I think since we are making this resolver smarter now, I would go w/ 
1. log a trace about this fact
2. default to bytes

I can envision us making the defaulting behavior easily configurable in the future (I mean, of course they can use a custom resolver)

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
```diff
@@ -130,7 +130,11 @@ public <T> Schema<T> getSchema(Class<?> messageClass, boolean returnDefault) {
 
 	@Nullable
 	private Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
-		return this.customSchemaMappings.getOrDefault(messageClass, (returnDefault ? Schema.BYTES : null));
+		Schema<?> schema = this.customSchemaMappings.get(messageClass);
+		if (schema == null && returnDefault) {
+			return messageClass != null ? Schema.JSON(messageClass) : Schema.BYTES;
```
onobc, line 135: Wdyt about moving this defaulting picking logic into a protected method so that users could override just this piece of the resolver easily (custom resolver)? That would also be the place where we handle the "failed to do the Bar<T>..."

onobc (changes requested):
Looks good @cbornet. A couple of comments to address (or not) and then we can put this one in.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
cbornet: Sure

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
cbornet: Personal opinion: please refrain 😄 ! Using Lombok has a lot of implications (like imposing to use an IDE that supports it). So I would avoid it for as long as possible. There are not that many value objects and records can be used most of the time.

cbornet: Support for record in AVRO : https://github.com/apache/avro/pull/1680
Probably for version 1.12

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
```diff
@@ -129,8 +132,22 @@ public <T> Schema<T> getSchema(Class<?> messageClass, boolean returnDefault) {
...
-		return this.customSchemaMappings.getOrDefault(messageClass, (returnDefault ? Schema.BYTES : null));
+	protected Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
+		Schema<?> schema = this.customSchemaMappings.get(messageClass);
+		if (schema == null && returnDefault) {
+			if (messageClass != null) {
+				try {
+					return Schema.JSON(messageClass);
+				}
+				catch (Exception e) {
+					if (this.logger.isDebugEnabled()) {
```
onobc, line 143: This is the beauty of `LogAccessor`.... you can ditch the "if level enabled check" w/ 
```java
this.logger.debug(() -> "Failed to create JSON schema for %s".formatted(messageClass.getName());
```

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
```diff
@@ -129,8 +132,22 @@ public <T> Schema<T> getSchema(Class<?> messageClass, boolean returnDefault) {
 	}
 
 	@Nullable
-	private Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
-		return this.customSchemaMappings.getOrDefault(messageClass, (returnDefault ? Schema.BYTES : null));
+	protected Schema<?> getCustomSchemaOrMaybeDefault(Class<?> messageClass, boolean returnDefault) {
+		Schema<?> schema = this.customSchemaMappings.get(messageClass);
```
onobc, line 136: I wonder if we should make `customSchemaMappings` protected here so that custom impls could use it as well?

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarTemplateTests.java
onobc: >  please refrain 😄 ! Using Lombok has a lot of implications 
Yep, just venting mostly. If we do end up bringing it in it would only be at test dep level. But yeh, 2 of these manual pojos does not warrant it.

onobc (changes requested):
@cbornet - thx for the updates. looks good. I couple of minor comments to address.

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
cbornet: Child classes can already add/remove/getView. I'm not sure what else could be needed. Maybe this could be studied in a subsequent issue/PR ?

onobc: Yep, that should do. And good point - lets leave the "As a user I can easily extend the topic and schema resolver to suit my needs" story for another issue. On that note, please resolve (table) any of my related comments around extension on this and the other resolver PR.

onobc: https://github.com/spring-projects-experimental/spring-pulsar/issues/330

spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultSchemaResolver.java
cbornet: done

cbornet:
> do you mind taking a stroll through once the dust settles and update/add if needed/applicable?

Added some

onobc (approved):

PR #327, opened 2023-02-07
Pulsar Reader (imperative) Boot properties support

sobychacko:

### DIFF
--- a/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
+++ b/spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
@@ -79,6 +79,8 @@ public class PulsarProperties {
 
 	private final Admin admin = new Admin();
 
+	private final Reader reader = new Reader();
+
 	private final Defaults defaults = new Defaults();
 
 	public Consumer getConsumer() {
@@ -109,6 +111,10 @@ public class PulsarProperties {
 		return this.admin;
 	}
 
+	public Reader getReader() {
+		return this.reader;
+	}
+
 	public Defaults getDefaults() {
 		return this.defaults;
 	}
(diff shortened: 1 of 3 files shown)


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/PulsarProperties.java
```diff
@@ -2075,6 +2085,120 @@ public Map<String, Object> buildProperties() {
 
 	}
 
+	public static class Reader {
+
+		/**
+		 * Topic name.
```
onobc, line 2091: **[NIT]** there can be multiples

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarPropertiesTests.java
```diff
@@ -503,4 +504,38 @@ void functionProperties() {
...
 	}
 
+	@Nested
+	class ReaderPropertiesTests {
+
+		@Test
+		void readerProperties() {
+			Map<String, String> props = new HashMap<>();
+
+			props.put("spring.pulsar.reader.topic-names", "my-topic");
```
onobc, line 514: I noticed you use kebab-case here but camel case below props.

spring-pulsar-spring-boot-autoconfigure/src/test/java/org/springframework/pulsar/autoconfigure/PulsarPropertiesTests.java
```diff
@@ -511,7 +511,7 @@ class ReaderPropertiesTests {
 		void readerProperties() {
 			Map<String, String> props = new HashMap<>();
 
-			props.put("spring.pulsar.reader.topic-names", "my-topic");
```
onobc: Oh, sorry @sobychacko - I should have clarified. I meant that we typically use the relaxed named binding on the "left hand side" as that is typically what peepz will use in yml so it theoretically helps catch that.

onobc (approved):
LGTM @sobychacko !

PR #333, opened 2023-02-08
Add reader application properties doc

cbornet:


onobc (approved):
Thx for adding this @cbornet

PR #338, opened 2023-02-11
Add consumer/producer config props to binder and binding level

onobc:
Add full consumer/producer props to binder/binding props

* Re-use the consumer and producer props from PulsarProperties
* Add unit and integration test for newly added props


spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java
```diff
@@ -89,9 +86,7 @@ public Consumer<T> createConsumer(Schema<T> schema, @Nullable Collection<String>
 		if (!CollectionUtils.isEmpty(customizers)) {
 			customizers.forEach(customizer -> customizer.customize(consumerBuilder));
 		}
-		Consumer<T> consumer = consumerBuilder.subscribe();
-		this.consumers.add(consumer);
```
onobc, line 93: **[UNRELATED]** but a memory leak waiting to happen. I removed it as it was not being used by anything.

PR #339, opened 2023-02-13
Pulsar binder partition count defaults

sobychacko:
In Pulsar, partitioned topics are optional. If we set the default partition count to be 1, then it unnecessarily creates a partitoned topic. Therefore, we need to set the default partition count as 0 in the binder and extended binding properties.

### DIFF
--- a/spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarBinderConfigurationProperties.java
+++ b/spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarBinderConfigurationProperties.java
@@ -36,7 +36,7 @@ public class PulsarBinderConfigurationProperties {
 	@NestedConfigurationProperty
 	private final ProducerConfigProperties producer = new ProducerConfigProperties();
 
-	private int partitionCount = 1;
+	private int partitionCount = 0;
 
 	public ConsumerConfigProperties getConsumer() {
 		return this.consumer;
--- a/spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarBinderConfigurationProperties.java
+++ b/spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarBinderConfigurationProperties.java
@@ -18,6 +18,7 @@ package org.springframework.pulsar.spring.cloud.stream.binder.properties;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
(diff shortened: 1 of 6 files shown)


spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarConsumerProperties.java
```diff
@@ -41,7 +41,7 @@ public class PulsarConsumerProperties extends ConsumerConfigProperties {
 	@Nullable
 	private Class<?> messageValueType;
 
-	private Integer partitionCount = 1;
```
onobc, line 44: I would suggest we move this back to big integer (and change the property in the binder to big integer) as that allows us to differentiate "user has not set anything" from "this is the default". In this case, because we set a default it only provides consistency w/ the other fields that have moved over to the non-primitive primitive wrappers. 

Wdyt?

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarBinderConfigurationProperties.java
```diff
@@ -36,7 +36,7 @@ public class PulsarBinderConfigurationProperties {
 	@NestedConfigurationProperty
 	private final ProducerConfigProperties producer = new ProducerConfigProperties();
 
-	private int partitionCount = 1;
+	private int partitionCount = 0;
```
onobc, line 40: Not directly related to this PR, but Do you mind changing the accessor from `partitionCount` to `getPartitionCount` to keep consistent w/ the other config props.

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarConsumerProperties.java
```diff
@@ -41,7 +41,7 @@ public class PulsarConsumerProperties extends ConsumerConfigProperties {
...
+	private int partitionCount = 0;
```
onobc, line 45: Wdyt about using null for the default which we can then interpret as no partitions?

onobc (changes requested):
@sobychacko change looks good. I have a couple of nit comments to address.

spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/properties/PulsarConsumerProperties.java
sobychacko: ok, good point.

onobc (approved):
Thx for the updates @sobychacko . LGTM!

PR #350, opened 2023-02-15
Remove PulsarClientConfiguration

onobc:
* replace w/ map of config props


sobychacko:
@onobc, we probably need to clean up the docs about the removal of `PulsarClientConfiguration.` I think we added something about this in the docs.

onobc:
You must have missed that I pulled that out in the PR for service urls. I did not think we were taking it out but it was a bit of internal detail that would probably overwhelm the user when reading at first.

sobychacko (approved):

PR #351, opened 2023-02-16
Improve topic validation

### LINKED ISSUE #266: Partitioned Topics Work Once, Fail Subsequently
The createIfExists logic for partitioned topics will work the first time if it doesn't exist but subsequent times when it does exist it will fail.  I debugged the logic and topic names do not have `-parition-` in them as expected by this logic:
https://github.com/spring-projects-experimental/spring-pulsar/blob/main/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java#L146

Therefore it thinks the topic doesn't exist when it in fact does and it will try to create it and fail.

onobc:
* Check if topic exists already for both partitioned and unpartitioned cases prior to creating.

Fixes #266

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
@@ -178,7 +178,7 @@ public class PulsarAdministration
 					if (topic.isPartitioned()) {
 						if (existingTopicsInNamespace.contains(topicName)) {
 							throw new IllegalStateException(
-									"Topic '%s' already exists un-partitioned. Needs to be deleted first."
+									"Topic '%s' already exists un-partitioned - needs to be deleted first"
 											.formatted(topicName));
 						}
 						var matchingPartitions = getMatchingTopicPartitions(topic, existingTopicsInNamespace);
@@ -195,7 +195,7 @@ public class PulsarAdministration
 							}
 							else if (numberOfExistingPartitions > topic.numberOfPartitions()) {
 								throw new IllegalStateException(
-										"Topic '%s' found w/ %d partitions - can't shrink to %d. Needs to be deleted first"
+										"Topic '%s' found w/ %d partitions but can't shrink to %d - needs to be deleted first"
 												.formatted(topicName, numberOfExistingPartitions,
 														topic.numberOfPartitions()));
 							}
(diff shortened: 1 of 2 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -161,39 +164,52 @@ private void createOrModifyTopicsIfNeeded(Collection<PulsarTopic> topics) {
...
 			try {
-				List<String> existingTopicsInNamespace = admin.topics().getList(namespace);
+				var existingTopicsInNamespace = admin.topics().getList(namespace);
 
-				for (PulsarTopic topic : requestedTopics) {
+				for (var topic : requestedTopics) {
+					var topicName = topic.getFullyQualifiedTopicName();
 					if (topic.isPartitioned()) {
-						List<String> matchingPartitions = getMatchingTopicPartitions(topic, existingTopicsInNamespace);
+						if (existingTopicsInNamespace.contains(topicName)) {
```
onobc, line 179: This is part one of the change: "If there is already a non-partitioned topic out there - barf"

spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarAdministration.java
```diff
@@ -161,39 +164,52 @@ private void createOrModifyTopicsIfNeeded(Collection<PulsarTopic> topics) {
...
+										"Topic '%s' found w/ %d partitions - can't shrink to %d. Needs to be deleted first"
+												.formatted(topicName, numberOfExistingPartitions,
+														topic.numberOfPartitions()));
 							}
 						}
 					}
 					else {
-						if (!existingTopicsInNamespace.contains(topic.getFullyQualifiedTopicName())) {
-							this.logger.debug(() -> "Topic " + topic.getFullyQualifiedTopicName() + " does not exist.");
+						var matchingPartitions = getMatchingTopicPartitions(topic, existingTopicsInNamespace);
```
onobc, line 205: This is part two of the change: "If there is already a partitioned topic out there - barf"

spring-pulsar/src/test/java/org/springframework/pulsar/core/PulsarAdministrationTests.java
```diff
@@ -52,6 +52,7 @@
  */
 @ExtendWith(SpringExtension.class)
 @ContextConfiguration
+@SuppressWarnings("JUnitMalformedDeclaration")
```
onobc, line 55: This prevents the warnings from IDEA on commit saying that test methods should declare method parameters (which **IS** valid w/ junit5).

sobychacko (approved):

PR #356, opened 2023-02-17
Consolidate invokeHandler invocation

sobychacko:


spring-pulsar/src/main/java/org/springframework/pulsar/listener/adapter/PulsarMessagingMessageListenerAdapter.java
```diff
@@ -146,22 +145,9 @@ protected org.springframework.messaging.Message<?> toMessagingMessageFromReader(
 		return getMessageConverter().toMessageFromReader(record, reader, getType());
 	}
 
-	protected final Object invokeHandler(Object data, org.springframework.messaging.Message<?> message,
```
onobc, line 149: Basically the HandlerAdapter already supported the varargs and we were forcing this artficial constraint by permutations of different handle methods in the adapter. This is more flexible and lets us just pass the args and the handler adapter will find the proper handler, correct?

onobc (approved):
@sobychacko LGTM. 

One question, 
I take it that `PulsarRecordMessagingMessageListenerAdapter` was not being used by anything, correct?

spring-pulsar/src/main/java/org/springframework/pulsar/listener/adapter/PulsarMessagingMessageListenerAdapter.java
sobychacko: Yep, that is correct. We were artificially making that extra hierarchy.

PR #378, opened 2023-03-22
Better container properties sync

sobychacko:
* If consumer properties are directly overridden, we need to sync that with container properties. This commit tries to centralize this synching.

### DIFF
--- a/spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarExtendedBindingPropertiesTests.java
+++ b/spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarExtendedBindingPropertiesTests.java
@@ -24,6 +24,7 @@ import java.util.Map;
 
 import org.apache.pulsar.client.api.ProducerAccessMode;
 import org.apache.pulsar.client.api.SubscriptionMode;
+import org.apache.pulsar.client.api.SubscriptionType;
 import org.apache.pulsar.client.impl.conf.ConfigurationDataUtils;
 import org.apache.pulsar.client.impl.conf.ConsumerConfigurationData;
 import org.apache.pulsar.client.impl.conf.ProducerConfigurationData;
@@ -34,6 +35,7 @@ import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.boot.context.properties.source.ConfigurationPropertySource;
 import org.springframework.boot.context.properties.source.MapConfigurationPropertySource;
+import org.springframework.pulsar.listener.PulsarContainerProperties;
 import org.springframework.pulsar.spring.cloud.stream.binder.properties.PulsarExtendedBindingProperties;
 
 /**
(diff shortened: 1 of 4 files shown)


spring-pulsar-spring-cloud-stream-binder/src/test/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarExtendedBindingPropertiesTests.java
```diff
@@ -111,4 +113,33 @@ void consumerProperties() {
...
+		String[] topics = pulsarContainerProperties.getTopics();
+		assertThat(pulsarContainerProperties.getTopics()).isEmpty();
+		assertThat(pulsarContainerProperties.getTopicsPattern()).isNull();
+
+		pulsarContainerProperties.updateContainerProperties();
+
+		assertThat(pulsarContainerProperties.getSubscriptionName()).isEqualTo("my-foo-sbscription");
+		assertThat(pulsarContainerProperties.getSubscriptionType()).isEqualTo(SubscriptionType.Shared);
+		assertThat(pulsarContainerProperties.getTopics()).isNotEmpty().hasSize(1);
+		assertThat(pulsarContainerProperties.getTopics()).contains("my-topic");
```
onobc, line 141: **[SIMPLIFICATION]**

This 
```java
assertThat(pulsarContainerProperties.getTopics())
    containsExactly("my-topic");
```

is the same as 
```java
assertThat(pulsarContainerProperties.getTopics())
    .isNotEmpty()
    .hasSize(1)
    contains("my-topic");
```

spring-pulsar/src/main/java/org/springframework/pulsar/listener/PulsarContainerProperties.java
```diff
@@ -246,4 +257,35 @@ public void setPulsarConsumerProperties(Properties pulsarConsumerProperties) {
 		this.pulsarConsumerProperties = pulsarConsumerProperties;
 	}
 
+	@SuppressWarnings("unchecked")
+	public void updateContainerProperties() {
+		if (!this.pulsarConsumerProperties.isEmpty()) {
```
onobc, line 262: **[NIT]** An invert of this if and short-circuit return will allow all the code below to shift over.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/PulsarContainerProperties.java
```diff
@@ -246,4 +257,35 @@ public void setPulsarConsumerProperties(Properties pulsarConsumerProperties) {
...
+			if (this.pulsarConsumerProperties.containsKey(SUBSCRIPTION_NAME)) {
+				this.subscriptionName = (String) this.pulsarConsumerProperties.get(SUBSCRIPTION_NAME);
+			}
+			if (this.pulsarConsumerProperties.containsKey(SUBSCRIPTION_TYPE)) {
+				this.subscriptionType = (SubscriptionType) this.pulsarConsumerProperties.get(SUBSCRIPTION_TYPE);
+			}
+			if (this.pulsarConsumerProperties.containsKey(TOPIC_NAMES)) {
+				Object topicNamesFromConsumerProperties = this.pulsarConsumerProperties.get(TOPIC_NAMES);
```
onobc, line 270: These are set in code via
```java
		JavaUtils.INSTANCE.acceptIfNotNull(this.phase, instance::setPhase)
				.acceptIfNotNull(this.applicationContext, instance::setApplicationContext)
				.acceptIfNotNull(this.applicationEventPublisher, instance::setApplicationEventPublisher)
				.acceptIfNotNull(endpoint.getConsumerProperties(),
						instance.getContainerProperties()::setPulsarConsumerProperties);
```
from `org.springframework.pulsar.config.AbstractPulsarListenerContainerFactory#initializeContainer`. A step through debugger would tell what it actually is coming in as. This may only be coming in differently due to a test. If that is the case we could adjust the test instead.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/PulsarContainerProperties.java
```diff
@@ -246,4 +257,35 @@ public void setPulsarConsumerProperties(Properties pulsarConsumerProperties) {
 		this.pulsarConsumerProperties = pulsarConsumerProperties;
 	}
 
+	@SuppressWarnings("unchecked")
+	public void updateContainerProperties() {
+		if (!this.pulsarConsumerProperties.isEmpty()) {
+			if (this.pulsarConsumerProperties.containsKey(SUBSCRIPTION_NAME)) {
```
onobc, line 263: One way to reduce this boiler plate would be w/ something like this...
```java
applyPropIfSpecified(SUBSCRIPTION_NAME, this::setSubscriptionName);
```

```java
	private <T> void applyPropIfSpecified(String key, Consumer<T> setter) {
		if (this.pulsarConsumerProperties.containsKey(key)) {
			T value = (T) this.pulsarConsumerProperties.get(key);
			setter.accept(value);
		}
	}
```
Wdyt?

onobc (changes requested):
Looks good @sobychacko , I have added a few comments (nothing big).

PR #397, opened 2023-05-02
Use builder to autoconfigure PulsarProducerFactory

onobc:
- Apply PulsarProperties to producer builder rather than raw map of Pulsar properties


spring-pulsar-spring-boot-autoconfigure/src/main/java/org/springframework/pulsar/autoconfigure/ProducerConfigProperties.java
```diff
@@ -401,4 +403,41 @@ public Map<String, Object> buildProperties() {
 		return properties;
 	}
 
+	@SuppressWarnings("deprecation")
```
onobc, line 406: I had to leave the [Map:buildProperties()](https://github.com/spring-projects/spring-pulsar/pull/397/files#diff-2c1533a4e0f091eec304ab387e83ce24a15b3cb5477f1c2e0f1073f1ffe6690fL368) in the codebase to appease the layered props override code that the [SCSt Pulsar Binder](https://github.com/spring-projects/spring-pulsar/blob/0452a06e1106b04832e45ca7e9af5e84b2a90412/spring-pulsar-spring-cloud-stream-binder/src/main/java/org/springframework/pulsar/spring/cloud/stream/binder/PulsarMessageChannelBinder.java#L113-L120) uses. 

There was not a straightforward way to do the same thing using customizers. Need to brainstorm on this one a bit. For now, we just leave it in.

PR #403, opened 2023-05-08
Use shaded cache module for Reactive client

onobc:
- use `pulsar-client-reactive-producer-cache-caffeine-shaded`
- exclude `pulsar-client` from reactive client modules as it brings in unwanted transitive deps

### DIFF
--- a/spring-pulsar-reactive/build.gradle
+++ b/spring-pulsar-reactive/build.gradle
@@ -11,7 +11,7 @@ dependencies {
 		// spring-pulsar includes a pulsar-client with its unwanted transitive deps excluded
 		exclude group: "org.apache.pulsar", module: "pulsar-client"
 	}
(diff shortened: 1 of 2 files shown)


spring-pulsar-reactive/build.gradle
```diff
@@ -17,7 +25,6 @@ dependencies {
 	optional 'com.google.protobuf:protobuf-java'
 	optional 'com.jayway.jsonpath:json-path'
 	optional 'io.projectreactor:reactor-core'
-	optional 'org.apache.pulsar:pulsar-client-reactive-producer-cache-caffeine'
```
onobc, line 20: Now that Caffeine is shaded in the cache module we can bring it in by default as it is recommended to use the cache module. This is the same thing we do for the imperative side (it brings in shaded cache module as well).

PR #447, opened 2023-09-13
Move cache provider into proper package.

onobc:
The cache provider modules source files were previously packaged under the `spring.pulsar.core` name. 

This moves them to a package name that is specific to their module name. 

The motiivation of the change is to pre-emptively reduce confusion w/ the Java module system.


onobc:
cc: @philwebb

PR #453, opened 2023-09-20
Clarify dependency coordinates for 0.2.x

onobc:
Screen Shot 2023-09-20 at 13 06 11


axiopisty:
I think this update looks great and is really helpful. Thank you @onobc!

PR #455, opened 2023-09-21
Clarify dependency coordinates in documentation

axiopisty:

### DIFF
--- a/spring-pulsar-docs/src/main/asciidoc/project-state.adoc
+++ b/spring-pulsar-docs/src/main/asciidoc/project-state.adoc
@@ -5,6 +5,25 @@ The evolution of the project (by version) is as follows:
 
 [discrete]
 == Version 0.2.x
+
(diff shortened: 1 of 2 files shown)


onobc:
Thanks for the contribution @axiopisty - well done. I adjusted the locations a bit in this [follow on commit](https://github.com/spring-projects/spring-pulsar/commit/97d524eeaa9af316d4ba958c20633b2f88d02df8). Specifically,  to keep the exact dependency coordinates in the quick tour sections. I did call out the fact that they will change across versions in the "project status" intro. Also, in the quick tour I added the exact deps for 0.2x as you had done in your PR.  Take a peek and let me know what you think. We can always re-adjust later if needed. 



Screen Shot 2023-09-22 at 00 06 25

Screen Shot 2023-09-22 at 00 06 54
Screen Shot 2023-09-22 at 00 10 16

PR #462, opened 2023-09-25
Antora

rwinch:

### DIFF
--- a/gradle/update-copyrights.gradle
+++ b/gradle/update-copyrights.gradle
@@ -6,7 +6,7 @@ if (gitPresent) {
 
 ext {
 	if (gitPresent) {
-		modifiedFiles = files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') }
+		modifiedFiles = providers.provider { files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') } }
 	}
 }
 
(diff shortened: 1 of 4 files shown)


spring-pulsar-docs/src/main/antora/antora.yml
```diff
@@ -10,3 +10,7 @@ ext:
       local: true
     scan:
       dir: spring-pulsar-docs/build/generated-antora-resources
+asciidoc:
```
onobc, line 13: Ahhh ok. This is component version metadata vs. playbook global metadata, correct?

gradle/update-copyrights.gradle
```diff
@@ -6,7 +6,7 @@ if (gitPresent) {
 
 ext {
 	if (gitPresent) {
-		modifiedFiles = files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') }
+		modifiedFiles = providers.provider { files(grgit.status().unstaged.modified).filter{ f -> f.name.endsWith('.java') } }
```
onobc, line 9: This was in there previously but I removed as `gradle build` was failing on it. I thought maybe it was from Gradle 8 but did not look too much into it.

spring-pulsar-docs/src/main/antora/antora.yml
onobc: I just noticed your descriptions in the commit. Thanks for that. It answers this question.

PR #481, opened 2023-11-09
Deprecate (Reactive)PulsarListenerEndpointAdapter

vpavic:
This commit deprecates `PulsarListenerEndpointAdapter` and `ReactivePulsarListenerEndpointAdapter` in favor of default methods on `ListenerEndpoint` and its subinterfaces, which makes it a bit easier to provide custom `ListenerEndpoint` implementations.


onobc:
Thanks for the PR @vpavic ,

The listener hierarchy side of things is a bit messy currently and refactorings like these are much appreciated. That being said, we are currently in RC and will do a 1.0.0-GA on 2023-11-20. As such, the only changes we are considering at this moment are bug fixes. We can consider the proposal after that though and I am more than happy to review it sooner. 

Is the motivation of this change just a general improvement or it is unblocking you from doing something?

vpavic:
No worries, I understand the phase of the 1.0 is such that it's too late to accept this change immediately so close to the GA. There's always next minor release to target this at (that's also why I didn't refer any specific releases in the deprecation markers).

This isn't a blocker of any kind, just a convenience to have a simpler and more flexible listener hierarchy when implementing custom `ListenerEndpoint`s.

onobc:
Thanks for the contribution @vpavic !

Issue #480, opened 2023-11-09
Setting subscription name using `spring.pulsar.consumer.subscription.name` configuration property doesn't work


vpavic:
[Pulsar Listener section](https://docs.spring.io/spring-pulsar/docs/1.0.0-RC1/reference/reference/pulsar.html#pulsar-listener) of reference manual provides snippets that suggest it's possible to configure subscription name only using `spring.pulsar.consumer.subscription.name` configuration property and without having `subscriptionName` specified on `@PulsarListener`.

However, in practice, I don't see this being possible due to subscription name being generated in `PulsarListenerAnnotationBeanPostProcessor` if omitted on annotation:
https://github.com/spring-projects/spring-pulsar/blob/728a74bcb00c6ee5b884ea8641c48fd309240240/spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListenerAnnotationBeanPostProcessor.java#L364-L369

Which combined with `DefaultPulsarConsumerFactory` applying the generated subscription name to `ConsumerBuilder` results in the value specified by `spring.pulsar.consumer.subscription.name` configuration property being overwritten by generated one:
https://github.com/spring-projects/spring-pulsar/blob/728a74bcb00c6ee5b884ea8641c48fd309240240/spring-pulsar/src/main/java/org/springframework/pulsar/core/DefaultPulsarConsumerFactory.java#L86-L88

onobc:
Hi @vpavic - thanks for the report, I will take a look.

onobc:
Good catch @vpavic, you are correct. This does not work as advertised because (as you detailed in the description) the default is only applied if the incoming name is null and in this case the annotation processor has set it to the generated name.

That being said, I don't think it would be valuable to use the default sub name because it is likely there will be multiple `@PulsarListener`s which would then require separate names. What I think would be helpful would be to allow the setting a default listener container sub name generated prefix and then the name would be `MY_PREFIX(0...N)`. A more flexible step would be to introduce the concept of a subscription name generator and plug that into the listener/containers. 

For now I will remove the falsey out of the docs and keep this issue open for further exploration around this area. Sound good to you? 

Again, thanks for finding this and thanks for the detailed analysis.

vpavic:
Thanks for looking into this @onobc, I'll review the PR that updates the documentation.

[...]

PR #482, opened 2023-11-10
Clarify use of default subscription name in @PulsarListener

onobc:
See #480

cc: @vpavic

### DIFF
--- a/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar.adoc
+++ b/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar.adoc
@@ -151,7 +151,7 @@ To use `PulsarListener`, you need to use the `@EnablePulsar` annotation.
 When you use Spring Boot support, it automatically enables this annotation and configures all the components necessary for `PulsarListener`, such as the message listener infrastructure (which is responsible for creating the Pulsar consumer).
 `PulsarMessageListenerContainer` uses a `PulsarConsumerFactory` to create and manage the Pulsar consumer.
 
-Spring Boot auto-configuration also provides this consumer factory which you can further configure by specifying any of the {spring-boot-pulsar-config-props}[`spring.pulsar.consumer.*`] application properties.
+Spring Boot auto-configuration also provides this consumer factory which you can further configure by specifying most of the {spring-boot-pulsar-config-props}[`spring.pulsar.consumer.*`] application properties (`subscription-name` is ignored and is instead generated when not specified on the annotation).
 
 Let us revisit the `PulsarListener` code snippet we saw in the quick-tour section:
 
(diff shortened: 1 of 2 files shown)


spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar.adoc
```diff
@@ -151,7 +151,7 @@ To use `PulsarListener`, you need to use the `@EnablePulsar` annotation.
 When you use Spring Boot support, it automatically enables this annotation and configures all the components necessary for `PulsarListener`, such as the message listener infrastructure (which is responsible for creating the Pulsar consumer).
 `PulsarMessageListenerContainer` uses a `PulsarConsumerFactory` to create and manage the Pulsar consumer.
 
-Spring Boot auto-configuration also provides this consumer factory which you can further configure by specifying any of the {spring-boot-pulsar-config-props}[`spring.pulsar.consumer.*`] application properties.
+Spring Boot auto-configuration also provides this consumer factory which you can further configure by specifying most of the {spring-boot-pulsar-config-props}[`spring.pulsar.consumer.*`] application properties (`subscription-name` is ignored and is instead generated when not specified on the annotation).
```
vpavic, line 154: There's actually a typo (that was there before as well) in the property name - it should actually be `spring.pulsar.consumer.subscription.name` (instead of `.subscription-name`).

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar.adoc
```diff
@@ -173,17 +173,16 @@ public void listen(String message) {
...
 [source,yaml,indent=0,subs="verbatim"]
 ----
-spring.pulsar.consumer:
-  topics: hello-pulsar
-  subscription-name: hello-pulsar-subscription
+spring.pulsar.consumer.topics: hello-pulsar-topic
 ----
 
-NOTE: If the topic information is not directly provided, a <<topic-resolution-process-imperative,topic resolution process>> is used to determine the destination topic.
+When `subscription-name` is not provided an auto-generated subscription name will be used.
```
vpavic, line 183: Same typo as above.

Perhaps also worth being more explicit and saying _not provided on `@PulsarListener` annotation_?

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/reactive-pulsar.adoc
```diff
@@ -163,8 +163,7 @@ In this most basic form, you must still provide the topic name by setting the fo
 
 [source,yaml,indent=0,subs="verbatim"]
 ----
-spring.pulsar.reactive.consumer:
-  topic-names: hello-pulsar-topic
+spring.pulsar.reactive.consumer.topic-names: hello-pulsar-topic
```
vpavic, line 166: This should be `spring.pulsar.consumer.topics` - from what I see both imperative and reactive share the same configuration property for this.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar.adoc
onobc: Ahh yes, this is why we do these code review things :) 

This field used to be accurate before the move to Spring Boot where we synthesized the imperative and reactive properties together and simplified them. I will scan the doc in its entirety for any such changes.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/reactive-pulsar.adoc
onobc: Ditto from https://github.com/spring-projects/spring-pulsar/pull/482#discussion_r1389522710

PR #483, opened 2023-11-11
Add subscription name generator support

onobc:
This commit adds the ability to confgure a subscription name generator that will be used by `@PulsarListener` and `@ReactivePulsarListener` when no subscription name is specified.

See #480

Update to ref docs to follow in subsequent commit.

cc: @vpavic

Screen Shot 2023-11-10 at 23 54 55

Screen Shot 2023-11-11 at 13 18 22

### DIFF
--- a/spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/listener/ReactivePulsarListenerTestsBase.java
+++ b/spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/listener/ReactivePulsarListenerTestsBase.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2022-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.pulsar.reactive.listener;
+
+import java.util.List;
+
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.PulsarClientException;
+import org.apache.pulsar.client.api.SubscriptionInitialPosition;
+import org.apache.pulsar.reactive.client.adapter.AdaptedReactivePulsarClientFactory;
+import org.apache.pulsar.reactive.client.api.ReactivePulsarClient;
+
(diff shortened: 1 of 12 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarReaderAnnotationBeanPostProcessor.java
```diff
@@ -233,6 +233,11 @@ else if (startMessageIdString.equalsIgnoreCase("latest")) {
 		resolveReaderCustomizer(endpoint, pulsarReader);
 	}
 
+	private void resolveSubscriptionName(MethodPulsarReaderEndpoint<?> endpoint, PulsarReader pulsarReader) {
```
onobc, line 236: In going through this exercise I realized that we have `subscriptionName` on `@PulsarReader` which does not make sense for a `Reader`.  I am going to remove this from the Reader entirely in a subsequent commit. As such you will notice their is no new test coverage for the `@PulsarReader` since it will be removed soon.

spring-pulsar/src/main/java/org/springframework/pulsar/config/PulsarEndpointSubscriptionNameGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.springframework.pulsar.config;
+
+import java.util.Optional;
+
+/**
+ * Contract to generate a subscription name for the Pulsar consumer endpoints.
+ *
+ * @param <T> the annotation for the enpdoint
+ * ({@code PulsarListener, PulsarReader, ReactivePulsarListener})
```
onobc, line 25: I will remove PulsarReader from the places when I remove it later.

spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/listener/ReactivePulsarListenerTestsBase.java
```diff
@@ -0,0 +1,116 @@
...
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * Base implementation for {@link ReactivePulsarListener @ReactivePulsarListener} tests.
+ *
+ * @author Chris Bono
+ */
+@SpringJUnitConfig
+@DirtiesContext
+abstract class ReactivePulsarListenerTestsBase implements PulsarTestContainerSupport {
```
onobc, line 54: I took this opportunity to not add another 100 lines of code to the overly populated `ReactivePulsarListenerTests` and instead factored out this base and created a new subscription specific specialization of this. (same for imperative side)

onobc:
Based on @vpavic point in [this comment](https://github.com/spring-projects/spring-pulsar/issues/480#issuecomment-1808782380), adding a name generator is less valuable. We instead decided to go w/ an approach that if you have a single `@PulsarListener` and a single customizer registered then it will be automatically applied (same for `@PulsarReader` and `@ReactivePulsarListener`).

Closing in favor of https://github.com/spring-projects/spring-pulsar/commit/836d385d8696f28f1229ade22efd3fa6adf15720

PR #559, opened 2024-01-30
Add AUTO_CONSUME schema support

### LINKED ISSUE #380: Add support for AUTO_CONSUME schema type
Add support for using `schemaType = SchemaType.AUTO_CONSUME` on `@PulsarListener`.

onobc:
This PR adds support for AUTO_CONSUME schema type to:
- @PulsarListener
- @PulsarReader
- @ReactivePulsarListener

> [!NOTE]
> - [x] update docs

Resolves #380

### DIFF
--- a/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/config/MethodReactivePulsarListenerEndpoint.java
+++ b/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/config/MethodReactivePulsarListenerEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2023 the original author or authors.
+ * Copyright 2022-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,11 +26,11 @@ import org.apache.pulsar.client.api.Consumer;
 import org.apache.pulsar.client.api.DeadLetterPolicy;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.Messages;
-import org.apache.pulsar.client.api.Schema;
 import org.apache.pulsar.common.schema.SchemaType;
 
 import org.springframework.core.MethodParameter;
 import org.springframework.core.ResolvableType;
+import org.springframework.core.log.LogAccessor;
 import org.springframework.expression.BeanResolver;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.converter.SmartMessageConverter;
@@ -64,6 +64,8 @@ import reactor.core.publisher.Flux;
  * @author Chris Bono
  */
 public class MethodReactivePulsarListenerEndpoint<V> extends AbstractReactivePulsarListenerEndpoint<V> {
+
+	private final LogAccessor logger = new LogAccessor(this.getClass());
 
 	private Object bean;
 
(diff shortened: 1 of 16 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
```diff
@@ -140,15 +144,26 @@ protected AbstractPulsarMessageToSpringMessageAdapter<V> createMessageListener(
...
+		schemaResolver.resolveSchema(schemaType, messageType)
+			.ifResolvedOrElse(pulsarContainerProperties::setSchema,
+					(ex) -> this.logger
+						.warn(() -> "Failed to resolve schema for type %s - will default to BYTES (due to: %s)"
+							.formatted(schemaType, ex.getMessage())));
 		// Make sure the schemaType is updated to match the current schema
 		if (pulsarContainerProperties.getSchema() != null) {
-			SchemaType type = pulsarContainerProperties.getSchema().getSchemaInfo().getType();
-			pulsarContainerProperties.setSchemaType(type);
+			// AUTO_CONSUME schema info is null - but we know it should be AUTO_CONSUME
```
jonasgeiregat, line 159: Is it necessary to set the SchemaType again explicitly? As far as I can see it's derived from `pulsarContainerProperties` which come from the `PulsarListener` annotation, which should always be AUTO_CONSUME in this case. What am I missing?

onobc: This was originally here to deal w/ the fact that the input schema type to the resolver may result in a schema that is not necesarrilly the "same" as the type. Namely, the `SchemaType.NONE` is set on the listener which will result in a schema based on the input message type. 

However, in the special `AUTO_CONSUME` case it has no schemaInfo in it and I was just bypassing the logic but as you pointed out in this case it will be the proper SchemaType to Schema relation and I will adjust this accordingly and also add a better comment about why its here in the first place. 

Good catch! Thanks.

onobc: Additionally this chunk of logic really needs to be factored out into a common base or delegate as it is highly duplicated amongst the MethodPulsarListener/MethodPulsarReader/ReactiveMethodPulsarListener[Endpoint] classes.

spring-pulsar/src/test/java/org/springframework/pulsar/listener/PulsarListenerAutoConsumeSchemaTests.java
```diff
@@ -0,0 +1,191 @@
...
+import org.springframework.test.context.ContextConfiguration;
+
+/**
+ * Tests for {@link PulsarListener @PulsarListener} using {@code schemaType} of
+ * {@link SchemaType#AUTO_CONSUME}.
+ *
+ * @author Chris Bono
+ */
+@ContextConfiguration(classes = PulsarListenerAutoConsumeSchemaTestsConfig.class)
+class PulsarListenerAutoConsumeSchemaTests extends PulsarListenerTestsBase {
```
onobc, line 59: The "having to test the same exact thing in 3 different places" is causing duplication headache. 

The problem comes down to the fact that the difference in each one is static in nature, the `@PulsarListener @PulsarReader @ReactivePulsarListener` is the slight variant amongst each of these. I started to go down the path of figuring out a way to abstract this out and my "time box" ran out of time for now. 

I will play w/ this at some point in the near future to reduce this down as this violation of DRY is highly annoying to me.

spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/config/MethodReactivePulsarListenerEndpoint.java
```diff
@@ -134,18 +136,29 @@ protected AbstractPulsarMessageToSpringMessageAdapter<V> createMessageHandler(
...
 		schemaResolver.resolveSchema(schemaType, messageType)
-			.ifResolved(schema -> pulsarContainerProperties.setSchema((Schema) schema));
-
-		// Make sure the schemaType is updated to match the current schema
+			.ifResolvedOrElse(pulsarContainerProperties::setSchema,
+					(ex) -> this.logger
+						.warn(() -> "Failed to resolve schema for type %s - will default to BYTES (due to: %s)"
+							.formatted(schemaType, ex.getMessage())));
+
+		// Attempt to make sure the schemaType is updated to match the resolved schema.
```
onobc, line 153: @jonasgeiregat I ended up doing a bit of squashing but this is the area you requested to simplify. I did that as well as added a comment about "Why".

spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
```diff
@@ -140,15 +143,26 @@ protected AbstractPulsarMessageToSpringMessageAdapter<V> createMessageListener(
...
-		schemaResolver.resolveSchema(schemaType, messageType).ifResolved(pulsarContainerProperties::setSchema);
-
-		// Make sure the schemaType is updated to match the current schema
+		schemaResolver.resolveSchema(schemaType, messageType)
+			.ifResolvedOrElse(pulsarContainerProperties::setSchema,
+					(ex) -> this.logger
+						.warn(() -> "Failed to resolve schema for type %s - will default to BYTES (due to: %s)"
+							.formatted(schemaType, ex.getMessage())));
+
+		// Attempt to make sure the schemaType is updated to match the resolved schema.
```
onobc, line 157: @jonasgeiregat [ditto](https://github.com/spring-projects/spring-pulsar/pull/559/commits/1eccb84d3d714b0d78be75ff3dc08d172dfcd28d#r1473797230)

spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarReaderEndpoint.java
```diff
@@ -114,21 +117,30 @@ protected AbstractPulsarMessageToSpringMessageAdapter<V> createReaderListener(
...
-		schemaResolver.resolveSchema(schemaType, messageType).ifResolved(pulsarContainerProperties::setSchema);
-
-		// Make sure the schemaType is updated to match the current schema
+		schemaResolver.resolveSchema(schemaType, messageType)
+			.ifResolvedOrElse(pulsarContainerProperties::setSchema,
+					(ex) -> this.logger
+						.warn(() -> "Failed to resolve schema for type %s - will default to BYTES (due to: %s)"
+							.formatted(schemaType, ex.getMessage())));
+
+		// Attempt to make sure the schemaType is updated to match the resolved schema.
```
onobc, line 131: @jonasgeiregat [ditto](https://github.com/spring-projects/spring-pulsar/pull/559/commits/1eccb84d3d714b0d78be75ff3dc08d172dfcd28d#r1473797230)

spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
onobc: Updated in each of the MethodReactive/PulsarListener/ReaderEndpoints.

jonasgeiregat (approved):

PR #572, opened 2024-02-11
Add support for AUTO_PRODUCE schema

onobc:
This commit adds support for producing raw JSON or Avro payloads with AUTO_SCHEMA.

### DIFF
--- a/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/attributes/attributes-variables.adoc
+++ b/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/attributes/attributes-variables.adoc
@@ -14,7 +14,7 @@
 :spring-cloud-stream-docs: https://docs.spring.io/spring-cloud-stream/docs/{spring-cloud-stream-version}/reference/html/
 :spring-cloud-function: https://spring.io/projects/spring-cloud-function
 
-:apache-pulsar-docs: https://pulsar.apache.org/docs/3.1.x
+:apache-pulsar-docs: https://pulsar.apache.org/docs/3.2.x
 :apache-pulsar-cient-docs: {apache-pulsar-docs}/client-libraries-java
 :apache-pulsar-io-docs: {apache-pulsar-docs}/io-connectors
 :apache-pulsar-function-docs: {apache-pulsar-docs}/functions-overview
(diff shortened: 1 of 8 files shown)


spring-pulsar-docs/src/main/antora/modules/ROOT/pages/attributes/attributes-variables.adoc
```diff
@@ -14,7 +14,7 @@
 :spring-cloud-stream-docs: https://docs.spring.io/spring-cloud-stream/docs/{spring-cloud-stream-version}/reference/html/
 :spring-cloud-function: https://spring.io/projects/spring-cloud-function
 
-:apache-pulsar-docs: https://pulsar.apache.org/docs/3.1.x
+:apache-pulsar-docs: https://pulsar.apache.org/docs/3.2.x
```
onobc, line 17: **[UNRELATED]** Bump in docs to current version

spring-pulsar/src/main/java/org/springframework/pulsar/core/CachingPulsarProducerFactory.java
```diff
@@ -193,7 +197,8 @@ static class ProducerCacheKey<T> {
 			Assert.notNull(schema, () -> "'schema' must be non-null");
 			Assert.notNull(topic, () -> "'topic' must be non-null");
 			this.schema = schema;
-			this.schemaHash = SchemaHash.of(this.schema);
+			this.schemaHash = (schema instanceof AutoProduceBytesSchema) ? AUTO_PRODUCE_SCHEMA_HASH
```
onobc, line 200: This is necessary as any call to `AutoProduceSchema.getSchemaInfo()` will throw RuntimeException on the client until after something has been sent and validated w/ the target topic (i.e a schema is initialized on it). In our case, we don't care about schema info, the producer can be used for any `byte[]` messages that the user wants to be validated against the target topic schema and this cache key will do that.

spring-pulsar/src/test/java/org/springframework/pulsar/core/CachingPulsarProducerFactoryTests.java
```diff
@@ -74,7 +75,7 @@ void cleanupFromTests() {
 	}
 
 	@Test
-	void createProducerMultipleCalls() throws PulsarClientException {
```
onobc, line 77: **[UNRELATED]** Leftover from removal of checked exception in Spring Pulsar APIs.

PR #592, opened 2024-02-26
Polish "Add consumer test utility"

onobc:
@jonas-grgt amazing work on the [PulsarConsumerTestUtil](https://github.com/spring-projects/spring-pulsar/commit/5bf9ba09446331150af557ccab8455035e3afc85) 🎉 

I was doing normal "polish" commit and I ended up making a few changes that I wanted to run by you rather than just piggyback in. 

### Nit polish things
- update javadocs a bit
- reorder some constructors based on specificity
- rename the test to end w/ "Tests" (plural) for consistency
- simplified doc examples a bit w/ String message type

### Changes
- add null check to "withSchema"
- drop the exact match API (more detail in comment inline)
- rename the conditions factory methods (more detail in comment inline)

Some of these things I missed during code review and some of them I just did not see until I pulled the util down and started to use it etc..


spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/testing-applications.adoc
```diff
@@ -1,42 +1,46 @@
 [[testing-applications]]
 = Testing Applications
+
```
onobc, line 3: When there is a blank line it prevents it from showing the TOC. 
Once we publish the module then we can link this awesome document you added to the navbar.adoc.

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/ConsumedMessagesConditions.java
```diff
@@ -41,42 +42,30 @@ static <T> ConsumedMessagesCondition<T> desiredMessageCount(int messageCount) {
...
-	 * value.
+	 * Verifies that the expected value equals the message payload value of at least one
+	 * consumed message.
 	 * @param expectation the expected value
 	 * @param <T> the type of the message
 	 * @return the condition
 	 */
-	static <T> ConsumedMessagesCondition<T> anyMessageMatchesExpected(T expectation) {
-		return messages -> messages.stream().anyMatch(message -> message.getValue().equals(expectation));
+	static <T> ConsumedMessagesCondition<T> atLeastOneMessageMatches(T expectation) {
```
onobc, line 51: The naming of these I struggled with until I started using the util. 
> "consume messages until at least one message matches 'a'" 

helps me understand what this is doing. 

Likewise w/ the N variant below, 
> "consume messages until at least one message matches each of 'a', 'b', and 'c'"

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/ConsumedMessagesConditions.java
```diff
@@ -41,42 +42,30 @@ static <T> ConsumedMessagesCondition<T> desiredMessageCount(int messageCount) {
...
 
-	/**
-	 * Verifies that the consumed messages value contains exactly the expected values.
-	 * @param expectation the expected values
-	 * @param <T> the type of the message
-	 * @return the condition
-	 */
-	@SafeVarargs
-	@SuppressWarnings("varargs")
-	static <T> ConsumedMessagesCondition<T> containsExactlyExpectedValues(T... expectation) {
```
onobc, line 77: I am proposing to drop this variant as I think it may be a bit confusing as well as it will keep looking for matches up to the timeout even though we know that once it passes N messages w/o a match we could short circuit (i.e. it will never match). 

Also, I would like to avoid maintaining an ever growing list of condition permutations. I propose we go out w/ these 3 and then per user request add new conditions when people ask for them. Sound good?

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -67,6 +69,13 @@ public SchemaSpec<T> fromTopic(String topic) {
 		return this;
 	}
 
+	@Override
+	public ConditionsSpec<T> withSchema(Schema<T> schema) {
+		Assert.notNull(schema, "Schema must not be null");
```
onobc, line 74: Moved this up in order-must-be-executed for a nice readable mental flow and also added null check. Did you leave it out for a particular reason that I am missing?

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -80,17 +89,11 @@ public ConditionsSpec<T> until(ConsumedMessagesCondition<T> condition) {
...
-		this.schema = schema;
-		return this;
-	}
-
 	@Override
 	public List<Message<T>> get() {
 		var messages = new ArrayList<Message<T>>();
 		try {
-			String subscriptionName = UUID.randomUUID() + "-test-consumer";
+			var subscriptionName = "test-consumer-%s".formatted(UUID.randomUUID());
```
onobc, line 96: Swapped to suffix as it will be easier to debug later "test-consumer-lfkdlsajskdljdfsa" than "lfjdslfjfadl-test-consumer" (probably)

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/ConsumedMessagesConditionsTests.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Tests for {@link ConsumedMessagesConditions}.
+ *
+ * @author Jonas Geiregat
+ */
+class ConsumedMessagesConditionsTests {
```
onobc, line 41: I'm sorry these are showing up as delete/add rather than a rename 😿 

All I did was collapse some spacing and re-order a bit. These tests you did are solid.

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @author Jonas Geiregat
+ */
+class PulsarConsumerTestUtilTests implements PulsarTestContainerSupport {
+
+	private PulsarTemplate<Object> pulsarTemplate;
+
+	private DefaultPulsarConsumerFactory<String> pulsarConsumerFactory;
+
+	private static String testTopic(String suffix) {
```
onobc, line 50: Same as [above](https://github.com/spring-projects/spring-pulsar/pull/592/files#r1502001480) 😿 

- I also scoped the topic name a bit more w/ prefix of acronym of the test class name (we do this in other tests as well). 
- I also used the more specific AssertJ "assertThat(IllegalArgument|IllegalState|ExceptionOfType)` APIs

Issue #561, opened 2024-02-01
Flaky ReactivePulsarListenerTombstoneTests$SingleComplexPayload causing random CI PR failures


onobc:
The `ReactivePulsarListenerTombstoneTests$SingleComplexPayload` fails randomly from time/time.

Example: https://github.com/spring-projects/spring-pulsar/actions/runs/7735972858/job/21092530057#step:4:7079

This needs to be fixed.

KartikShrivastava:
Hi @onobc, can i work on this?

onobc:
Hi @KartikShrivastava ,

That would be great. The trick will be to reproduce it. I have not had luck doing so locally. Let me know if you have any questions.

KartikShrivastava:
Hi @onobc,
I too couldn't reproduce it locally. I have two hypothesis tho:
- Messages are received in order different then the test expects
- `assertMessagesReceivedWithHeaders` is called before `latchWithHeaders` is finished executing?

wdyt?

P.S.:
Linking the CI PRs failures I found associated with this flakiness [[1]](https://github.com/spring-projects/spring-pulsar/actions/runs/7735972858/job/21092530057) [[2]](https://github.com/spring-projects/spring-pulsar/actions/runs/7598761030/job/20695067626)

onobc:
Hi @KartikShrivastava ,
Thanks for digging into this! 

Those are very plausible theories. I am not opposed to adding temporary code in tests or main (logging etc..) that may help prove the points. Do you see anything (and anywhere) we can instrument to see if we can get some more info during the CI run?

onobc:
Hi @KartikShrivastava ,
If you do not want to continue looking into this issue I totally understand. The issues currently in the backlog are rather coarse grained and if you are looking something small to work on there is not a lot to choose from. We have been meaning to put some smaller issues in and mark them accordingly. I will try to get to that this week but in the meantime if you want me to find you something else just let me know. And there is no obligation to do anything - I just wanted to offer this in case. 

Thanks

KartikShrivastava:
Hi @onobc, apologies for not raising that small PR yet, I was planning to do it in coming weekend. My weekdays availability is most of the time very less.

P.S: Please feel free to assign any other small and low priority task you have in mind, I find this project quite interesting to explore and learn from

onobc:
[...]

PR #597, opened 2024-03-03
Add temporary info logs to analyse flaky test issue #561

KartikShrivastava:
Due to flakiness of `ReactivePulsarListenerTombstoneTests$SingleComplexPayload` test, CI PR worflow has been failing intermetently. The exact reason of this flaky behaviour is not known so this PR adds an info log to give more insights on state of `receivedMessagesWithHeaders` before it is asserted.
- Reference issue: https://github.com/spring-projects/spring-pulsar/issues/561

### What does it do?
- Updates `spring-pulsar-reactive/src/test/resources/logback-test.xml` by change log level of `ReactivePulsarListenerTombstoneTests$SingleComplexPayload` to `info`
- Adds an info log in `ReactivePulsarListenerTombstoneTests$SingleComplexPayload$shouldReceiveMessagesWithTombstone` test

### DIFF
--- a/spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/listener/ReactivePulsarListenerTombstoneTests.java
+++ b/spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/listener/ReactivePulsarListenerTombstoneTests.java
@@ -33,6 +33,7 @@ import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.log.LogAccessor;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.handler.annotation.Header;
 import org.springframework.messaging.handler.annotation.Payload;
@@ -226,6 +227,8 @@ class ReactivePulsarListenerTombstoneTests extends ReactivePulsarListenerTestsBa
 	@ContextConfiguration(classes = SingleComplexPayloadConfig.class)
 	class SingleComplexPayload {
 
+		private final LogAccessor logger = new LogAccessor(this.getClass());
+
 		private static final String TOPIC = "rpltt-single-complex-topic";
 
 		static CountDownLatch latchWithHeaders = new CountDownLatch(3);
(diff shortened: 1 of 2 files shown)


spring-pulsar-reactive/src/test/java/org/springframework/pulsar/reactive/listener/ReactivePulsarListenerTombstoneTests.java
```diff
@@ -240,6 +243,13 @@ void shouldReceiveMessagesWithTombstone() throws Exception {
 			sendTestMessages(fooPulsarTemplate, TOPIC, Schema.JSON(Foo.class), Foo::new);
 			assertThat(latchWithHeaders.await(10, TimeUnit.SECONDS)).isTrue();
 			assertThat(latchWithoutHeaders.await(10, TimeUnit.SECONDS)).isTrue();
+
+			// Temporary log to analyze CI failures due to flaky test, one such failure
+			// case:
+			// https://github.com/spring-projects/spring-pulsar/actions/runs/7598761030/job/20695067626
+			for (ReceivedMessage<Foo> message : receivedMessagesWithHeaders) {
+				logger.info(message.toString());
+			}
```
onobc, line 252: HI @KartikShrivastava ,
I think this will be helpful. I would adjust this slightly to add the TODO for removal and remove the link from the comment (we have it in the issue). Also, another thing to help find the needle in the haystack (the log output :) ) is to prefix the message w/ some well-known string such as: 

```suggestion
			// Temporary log to analyze CI failures due to flaky test
			// TODO: Remove once CI failure addressed
			logger.info(() -> "***** TombstoneTestDebug: %s".formatted(receivedMessagesWithHeaders));
```

Wdyt?

KartikShrivastava: > I think this will be helpful. I would adjust this slightly to add the TODO for removal and remove the link from the comment (we have it in the issue). 

+1 for this, and I've made the changes

> Also, another thing to help find the needle in the haystack (the log output :) ) is to prefix the message w/ some well-known string

That's a valid point but I haven't added any extra prefix because `LogAccessor` also prefix the log with the path to the method along with other identifier which we can use, here's what it looks like right now in the CI:
```
ReactivePulsarListenerTombstoneTests > SingleComplexPayload > shouldReceiveMessagesWithTombstone() STANDARD_OUT
    00:37:37.659 [Test worker] INFO  org.springframework.pulsar.reactive.listener.ReactivePulsarListenerTombstoneTests$SingleComplexPayload - ReceivedMessage[payload=Foo[value=foo], keyHeader=key:foo]
    00:37:37.659 [Test worker] INFO  org.springframework.pulsar.reactive.listener.ReactivePulsarListenerTombstoneTests$SingleComplexPayload - ReceivedMessage[payload=null, keyHeader=key:null]
    00:37:37.659 [Test worker] INFO  org.springframework.pulsar.reactive.listener.ReactivePulsarListenerTombstoneTests$SingleComplexPayload - ReceivedMessage[payload=Foo[value=bar], keyHeader=key:bar]
```
Hope that will work?

onobc:
Thanks @KartikShrivastava ,
It is merged - now we just have to sit and wait for it to fail again lol.

PR #601, opened 2024-03-10
Enhance consumer test util

### LINKED ISSUE #599: Enhance ConsumerTestUtil
### Convenience method variants for `consumeMessages`
Currently the `PulsarConsumerTestUtil` requires the `consumeMessages` to pass in a consumer factory. Adding the following convenience variants would make using it simpler:

- `consumeMessages()`
Create the PulsarClient and PulsarConsumerFactory by first trying "localhost:6650" and if that fails, look at the 
`PulsarTestContainerSupport` coordinates.

- `consumeMessages(String url)`
Create the PulsarClient and PulsarConsumerFactory given the server url.

- ` consumeMessages(PulsarClient)`
Create the PulsarConsumerFactory given the PulsarClient

### Do not call `until` more than once
If user calls `until` > 1 then we should probably throw exception w/ message telling them to instead use `and` on their conditions.

KartikShrivastava:
Resolves #599

### DIFF
--- a/spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
+++ b/spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
@@ -55,7 +55,7 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
 
 	private List<String> topics;
 
-	private boolean calledUntil = false;
+	private boolean untilMethodAlreadyCalled = false;
 
 	public static <T> TopicSpec<T> consumeMessages() {
 		if (!PulsarTestContainerSupport.isContainerStarted()) {
@@ -111,11 +111,11 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
 
 	@Override
 	public ConditionsSpec<T> until(ConsumedMessagesCondition<T> condition) {
-		if (calledUntil) {
+		if (untilMethodAlreadyCalled) {
 			throw new IllegalStateException(
 					"Multiple calls to 'until' are not allowed. Use 'and' to combine conditions.");
 		}
-		this.calledUntil = true;
+		this.untilMethodAlreadyCalled = true;
 		this.condition = condition;
 		return this;
 	}
--- a/spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
+++ b/spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
@@ -60,33 +60,23 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
 	public static <T> TopicSpec<T> consumeMessages() throws PulsarClientException {
 		PulsarClient pulsarClient;
 		try {
-			pulsarClient = PulsarClient.builder().serviceUrl(
-				"pulsar://localhost:6650"
-			).build();
+			pulsarClient = PulsarClient.builder().serviceUrl("pulsar://localhost:6650").build();
 		}
 		catch (PulsarClientException e) {
-			pulsarClient = PulsarClient.builder().serviceUrl(
-				PulsarTestContainerSupport.getPulsarBrokerUrl()
-			).build();
+			pulsarClient = PulsarClient.builder().serviceUrl(PulsarTestContainerSupport.getPulsarBrokerUrl()).build();
 		}
-		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(
-			pulsarClient, List.of()
-		);
+		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(pulsarClient, List.of());
 		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
 	}
 
 	public static <T> TopicSpec<T> consumeMessages(String url) throws PulsarClientException {
 		PulsarClient pulsarClient = PulsarClient.builder().serviceUrl(url).build();
-		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(
-			pulsarClient, List.of()
-		);
+		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(pulsarClient, List.of());
 		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
 	}
 
 	public static <T> TopicSpec<T> consumeMessages(PulsarClient pulsarClient) {
-		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(
-			pulsarClient, List.of()
-		);
+		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(pulsarClient, List.of());
 		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
 	}
 
@@ -124,8 +114,7 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
 	public ConditionsSpec<T> until(ConsumedMessagesCondition<T> condition) {
 		if (calledUntil) {
 			throw new IllegalStateException(
-				"Multiple calls to 'until' are not allowed. Use 'and' to combine conditions."
-			);
+					"Multiple calls to 'until' are not allowed. Use 'and' to combine conditions.");
 		}
 		this.calledUntil = true;
 		this.condition = condition;
--- a/spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
+++ b/spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
@@ -57,27 +57,26 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
 
 	private boolean calledUntil = false;
 
-	public static <T> TopicSpec<T> consumeMessages() throws PulsarClientException {
-		PulsarClient pulsarClient;
-		try {
-			pulsarClient = PulsarClient.builder().serviceUrl("pulsar://localhost:6650").build();
+	public static <T> TopicSpec<T> consumeMessages() {
+		if (!PulsarTestContainerSupport.isContainerStarted()) {
+			return PulsarConsumerTestUtil.consumeMessages(PulsarTestContainerSupport.getPulsarBrokerUrl());
 		}
-		catch (PulsarClientException e) {
-			pulsarClient = PulsarClient.builder().serviceUrl(PulsarTestContainerSupport.getPulsarBrokerUrl()).build();
-		}
-		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(pulsarClient, List.of());
-		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
+		return PulsarConsumerTestUtil.consumeMessages("pulsar://localhost:6650");
 	}
 
-	public static <T> TopicSpec<T> consumeMessages(String url) throws PulsarClientException {
-		PulsarClient pulsarClient = PulsarClient.builder().serviceUrl(url).build();
-		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(pulsarClient, List.of());
-		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
+	public static <T> TopicSpec<T> consumeMessages(String url) {
+		Assert.notNull(url, "url must not be null");
+		try {
+			return PulsarConsumerTestUtil.consumeMessages(PulsarClient.builder().serviceUrl(url).build());
+		}
+		catch (PulsarClientException ex) {
+			throw new PulsarException(ex);
+		}
 	}
 
 	public static <T> TopicSpec<T> consumeMessages(PulsarClient pulsarClient) {
-		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(pulsarClient, List.of());
-		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
+		Assert.notNull(pulsarClient, "pulsarClient must not be null");
+		return PulsarConsumerTestUtil.consumeMessages(new DefaultPulsarConsumerFactory<>(pulsarClient, List.of()));
 	}
 
 	public static <T> TopicSpec<T> consumeMessages(PulsarConsumerFactory<T> pulsarConsumerFactory) {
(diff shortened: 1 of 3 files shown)


spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -53,6 +55,41 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
...
+	public static <T> TopicSpec<T> consumeMessages() throws PulsarClientException {
+		PulsarClient pulsarClient;
+		try {
+			pulsarClient = PulsarClient.builder().serviceUrl(
+				"pulsar://localhost:6650"
+			).build();
+		}
+		catch (PulsarClientException e) {
+			pulsarClient = PulsarClient.builder().serviceUrl(
+				PulsarTestContainerSupport.getPulsarBrokerUrl()
```
jonas-grgt, line 69: What if the test using `consumeMessages()` does not rely on`PulsarTestContainerSupport`? In that case , I believe, the pulsar container will not have been started. Maybe we should verify if the `PulsarTestContainerSupport` container is actually being used (started) and if not throw a proper exception?
Or even better, this could be the first check at the start of the method? Verify if the container is started, if so use its `serviceUrl` else default to `localhost:6650`?

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -85,6 +122,12 @@ public ConditionsSpec<T> awaitAtMost(Duration timeout) {
 
 	@Override
 	public ConditionsSpec<T> until(ConsumedMessagesCondition<T> condition) {
+		if (calledUntil) {
```
jonas-grgt, line 125: Do we really need to introduce a new variable? Is there a reason why we can't rely on `this.condition` being not being `null` to verify if `until` has been called before?

jonas-grgt (changes requested):
Wonderful to see someone pick up and improve spring-pulsar based on the work I did a couple of weeks back! Way to go @KartikShrivastava  🎉

You did a really nice job and it's definitely a worthwhile contribution. I did raise a couple of questions. 

There are also some formatting violations that are easy to fix simply by running `./gradlew format`

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
onobc: Good point @jonas-grgt , I think flipping them makes sense. 

Also, it would be helpful to add the following to PTCS to ease this check.

```java
static boolean isContainerStarted() {
	return PULSAR_CONTAINER.isRunning();
}
```

With this in place we can heavily assume that if the PTCS is running then this is the only way we should try to connect. It would be confusing if it was running but for some reason it failed to connect and then we try `localhost`. 

Something like this: 

```java
public static <T> TopicSpec<T> consumeMessages() {
	if (!PulsarTestContainerSupport.isContainerStarted()) {
		return PulsarConsumerTestUtil.consumeMessages(PulsarTestContainerSupport.getPulsarBrokerUrl());
	}
	return PulsarConsumerTestUtil.consumeMessages("pulsar://localhost:6650");
}

```

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
onobc: I believe we do since we do not have to specify a condition (e.g. `PCTUT.whenConditionIsNotSpecifiedMessagesAreConsumedUntilAwaitDuration`). 

I would like to suggest to rename the var to `untilMethodAlreadyCalled` as the word `until` in general is a confusing one that makes my brain pause when reading the code "until something happens". It seems nit picky but including the "method" in there helps readability for me anyways.

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
```diff
@@ -97,9 +115,27 @@ void exceptionIsThrownWhenConditionNotMetWithinAwaitDuration() {
 			.withMessage("Condition was not met within 5 seconds");
 	}
 
+	@Test
+	void exceptionIsThrownWhenMultipleUntilConditionsAreChainedTogether() {
```
onobc, line 119: **[TinyNit]** I believe an exception is thrown when until is called multiple times.

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -53,6 +55,41 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
...
+				PulsarTestContainerSupport.getPulsarBrokerUrl()
+			).build();
+		}
+		PulsarConsumerFactory<T> pulsarConsumerFactory = new DefaultPulsarConsumerFactory<>(
+			pulsarClient, List.of()
+		);
+		return new PulsarConsumerTestUtil<>(pulsarConsumerFactory);
+	}
+
+	public static <T> TopicSpec<T> consumeMessages(String url) throws PulsarClientException {
```
onobc, line 78: This is great stuff @KartikShrivastava !

I am going to include a few points in this comment.

## Input validation
Let's add assert validate checks on each newly added method (consistent w/ the existing consumer factory variant.

## Checked exceptions
We (@jonas-grgt did all the work 😄 ) recently removed checked exceptions from our top level APIs. I think we should do the same here. The typical pattern is to `try/catch(PulsarClientException) -> throw PulsarException`. There is an example in one of the sample code blocks below. 

## Method chaining
I think we can chain these calls to consolidate them down a bit. E.g. the url variant can pass the client into the client variant etc. and we can do the same for the no-arg variant. Something like:

```java
public static <T> TopicSpec<T> consumeMessages(String url) {
	Assert.notNull(url, "url must not be null");
	try {
		return PulsarConsumerTestUtil.consumeMessages(PulsarClient.builder().serviceUrl(url).build());
	}
	catch (PulsarClientException ex) {
		throw new PulsarException(ex);
	}
}

public static <T> TopicSpec<T> consumeMessages(PulsarClient pulsarClient) {
	Assert.notNull(pulsarClient, "pulsarClient must not be null");
	var pulsarConsumerFactory = new DefaultPulsarConsumerFactory<T>(pulsarClient, List.of());
	return PulsarConsumerTestUtil.consumeMessages(pulsarConsumerFactory);
}
```
Wdyt?

onobc (changes requested):
@KartikShrivastava, I could not agree w/ @jonas-grgt more on this being a great contribution. Thank you for this. 

I also like that you submitted the WIP before investing too much into testing etc..that way we can align on current direction. Good call. 

I have left a few comments/suggestions. 

Great work.

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
KartikShrivastava: Sounds good to me, will update

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
KartikShrivastava: `untilMethodAlreadyCalled` is better than `calledUntil` imo

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -53,6 +55,30 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
 
 	private List<String> topics;
 
+	private boolean untilMethodAlreadyCalled = false;
+
+	public static <T> TopicSpec<T> consumeMessages() {
+		if (!PulsarTestContainerSupport.isContainerStarted()) {
```
onobc, line 61: We have the logic flipped here (need to remove the `!`).

onobc (changes requested):
@KartikShrivastava, thanks for the updates. LGTM. 

The only remaining piece is to add a few tests to cover the newly added methods.

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
KartikShrivastava: gotchu

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
```diff
@@ -97,9 +100,101 @@ void exceptionIsThrownWhenConditionNotMetWithinAwaitDuration() {
...
+			.until(desiredMessageCount(2))
+			.get();
+		assertThat(msgs).hasSize(2);
+	}
+
+	@Test
+	void messagesAreConsumedWhenContainerIsStoppedAndConsumeMessagesIsCalledWithoutArguments() {
+		PulsarTestContainerSupport.stopContainer();
+		PulsarConsumerTestUtil.<String>consumeMessages();
+		// TODO: Complete this test
```
KartikShrivastava, line 121: I'd like to request some assistance writing this test
- Basically I want to test if consumeMessages works properly even if `PULSAR_CONTAINER` is not running
- When I try to `get()` the consumeMessages I get connection refused error for the localhost pulsar client
- Also I'm not sure at what point should I send messages to the topic, if I do it before stopping the container will it still work? since the pulsarClient instance would be different after container is stopped

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
```diff
@@ -53,6 +55,30 @@ public class PulsarConsumerTestUtil<T> implements TopicSpec<T>, SchemaSpec<T>, C
...
+		if (PulsarTestContainerSupport.isContainerStarted()) {
+			return PulsarConsumerTestUtil.consumeMessages(PulsarTestContainerSupport.getPulsarBrokerUrl());
+		}
+		return PulsarConsumerTestUtil.consumeMessages("pulsar://localhost:6650");
+	}
+
+	public static <T> TopicSpec<T> consumeMessages(String url) {
+		Assert.notNull(url, "url must not be null");
+		try {
+			return PulsarConsumerTestUtil.consumeMessages(PulsarClient.builder().serviceUrl(url).build());
```
KartikShrivastava, line 70: I have a question around this piece of code:
- We are calling `build` on `PulsarClient` once again, so in situations when `PulsarClient` as already been built prior to calling this method would it cause any issues? Or this `build` method is idempotent?

KartikShrivastava:
Hey folks, I've added the tests for the changes made in `PulsarConsumerTestUtil` and left few questions I got while testing the new methods

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
onobc: The `PulsarClient.builder()` creates a new builder instance and will ultimately construct a new client instance. This is functionally fine but it does bring up a good point. In the cases where we are constructing the client and the consumer factory we need to close them once the util method completes. Let's add that in a subsequent PR. For now, this is fine.

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
onobc: Yep, this one is tricky to test. We actually can't stop the container as the other tests will fail. The "magic" of the PTCS is that it has the `@BeforeAll` that starts it for the test class. If we stopped it we would have to restart it and then it gets into a bit of flaky-test-land at that point. 

I think instead we can static mock the "isContainerStarted" that it uses to decide what to do. 

We are releasing `1.1.0-M2` tomorrow so I am going to take what you have in this current state and add a polish commit on top so that we can get this in. I will include this test adjustment in that commit. 

Great work @KartikShrivastava !

onobc: Also, if you are wondering why we did not have (or need) the stop container method annotated w/ a `@AfterAll` to match the start w/ / `@BeforeAll` it is because Testcontainers does

> At the end of the test suite the [Ryuk container](https://github.com/testcontainers/moby-ryuk) that is started by Testcontainers core will take care of stopping the singleton container.

However, I think I will add it in the polish commit to avoid future confusion

onobc:
Closing via https://github.com/spring-projects/spring-pulsar/commit/2c7e3b248a9ffacf482f5c63d7f018815ca15ef5

onobc:
@KartikShrivastava thanks for this great contribution. Take a peek at the follow on commit I did https://github.com/spring-projects/spring-pulsar/commit/d99101ff27e70c491b54633df16d5fc1c7ed9abb. You can see how I uses static mocks to test our difficult case. I re-ordered the test methods a bit as well. Great work. Thank you.

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
onobc: > [!NOTE]
> Adding the `@AfterAll` to the new `stopContainer` method had unintended side-effects when used in a Jupiter `@Nested` test. I removed the annotation but left the method (https://github.com/spring-projects/spring-pulsar/commit/9bc6bda306b347d47b7b7a5d44b4e56db5f81798)

spring-pulsar-test/src/main/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtil.java
KartikShrivastava: Cool, thanks for the explanation!

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
KartikShrivastava: Cool, thanks for taking this up @onobc, really appreciate that.
I've one minor question seeing the improvements you've made: Is there a recommended way or methodology we follow to name test methods in this project? If so, then I would like to learn/read more about it :)

KartikShrivastava:
Thanks @onobc!

spring-pulsar-test/src/test/java/org/springframework/pulsar/test/support/PulsarConsumerTestUtilTests.java
onobc: You are welcome @KartikShrivastava . 
In general we follow the "when-THIS-then-OUTCOME" but more than anything the principle I follow is consistency w/in a test suite and enough information in the name to understand what it is testing. I will also typically group tests together using `@Nested` which groups them and allows us to not have ultra long test names. I was tempted to do some grouping on this test but ended up leaving as is. 

I also noticed that the negative test cases I named a bit different than the other ones (they are shorter in nature). 

So there is no doc that describes this but it would be a good idea for us to add one. I will note this.

Thanks

Issue #27, opened 2022-07-16
Add support for transactions


onobc:
Add transaction support to the following components:

* `PulsarTemplate`
* `@PulsarListener`
* `DefaultPulsarMessageListenerContainer`

Also add `PulsarTemplate` local transactions via the `executeInTransaction` API.

onobc:
After recent commit the following items still need to be handled:

- [ ] add tests for `transaction.*` package 
- [ ] add docs
- [ ] add `@Transactional` tests for both template and listener
- [ ] AckUtils - call verification
- [ ] DPMLC - enforce no error handler w/ batch listener w/ txn

onobc:
Closed via d179573968228f36d592add5c85b699c267da610

PR #626, opened 2024-03-25
Add local transaction support to PulsarTemplate

onobc:
This commit allows users to operate on a template within the context of a thread bound native Pulsar transaction.

Also included is the foundation to support Spring platform transactions via the PulsarTransactionManager and supporting classes.

See #27

> [!NOTE]
> This is the 1st step to allow transaction support in the framework. The following items need to be done as well:
> - [ ] Add txn support for message consumption (`PulsarMessageListenerContainer`)
> - [ ] Add support for Spring platform txn (`@Transactional`)
> - [ ] Add docs

cc: @jonas-grgt 

I believe the big picture may not be visible in the scope of this PR but the next PR that updates the consumer side of things I believe will clear things up.
In general, the Spring platform txn mgr starts a binds a native Pulsar txn and invokes the commit/rollback at the proper time. Ultimately we need to update the template (send) and the listener (receive) to look for an active txn and if found set it on the operation accordingly.

### DIFF
--- a/spring-pulsar/spring-pulsar.gradle
+++ b/spring-pulsar/spring-pulsar.gradle
@@ -39,6 +39,9 @@ dependencies {
 	testImplementation 'org.junit.jupiter:junit-jupiter'
 	testImplementation 'org.mockito:mockito-junit-jupiter'
 	testImplementation 'org.springframework:spring-test'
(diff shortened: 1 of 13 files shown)


onobc:
Closing in favor of https://github.com/spring-projects/spring-pulsar/pull/628

Issue #463, opened 2023-09-27
[Docs] Split Pulsar reference into multiple documents


onobc:
The `pulsar.adoc` page is very long and renders as 26 pages in the page navigation view in Antora (which then requires scrolling). 
This should be split out into multiple pages where each section is a new page. 

See [spring-kafka](https://github.com/spring-projects/spring-kafka/tree/main/spring-kafka-docs/src/main/antora/modules/ROOT/pages/kafka) for a reference as it had the same issue.

KartikShrivastava:
Hey @onobc, can I work on this?

onobc:
Hi @KartikShrivastava ,

That would be most welcomed.

As a point of comparison you can look at the Spring Kafka docs https://docs.spring.io/spring-kafka/reference/ and you will see how the "Using Spring for Apache Kafka" https://docs.spring.io/spring-kafka/reference/kafka.html is split into sections. But if you look at the Spring for Apache Pulsar counterpart https://docs.spring.io/spring-pulsar/reference/reference/pulsar.html there is no sections and the right hand side of this page is a HUGE menu of nested links. It is hard to navigate. 

We want Spring Pulsar to look similar to Spring Kafka docs.

onobc:
Hi @KartikShrivastava ,
We have been busy and have not had a chance to review but should get to this w/in the next 24hrs. 

Thanks for the contribution.

PR #627, opened 2024-03-25
Splits Pulsar reference into multiple documents (Issue #463)

KartikShrivastava:
- This PR addresses the #463 issue
- The main theme is to improve the readability of `pulsar.adoc`

### What does it do
- This PR brings changes to Spring Pulsar ascii doc layout to improve readability
- Sections present in existing `pulsar.adoc` are moved to dedicated ascii docs
- Fixes anchor references after moving blocks to dedicated docs
- Note: `Preface` section has been kept in `pulsar.adoc` as I felt it would provide some overview to the reader in `pulsar.adoc` itself. But please let me know if it should also be a dedicated doc

### Screenshots
Before:
spring-pulsar-doc-before

After:
spring-pulsar-doc-after


spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/message-consumption.adoc
```diff
@@ -0,0 +1,1030 @@
...
+}
+----
+
+In the preceding example, the `PulsarListener` creates five different consumers (this time, we assume that the topic has five partitions).
+
+NOTE: In this version, there is no message ordering, as `Shared` subscriptions do not guarantee any message ordering in Pulsar.
+
+If you need message ordering and still want a shared subscription types, you need to use the `Key_Shared` subscription type.
+
+=== Message Consumption
```
onobc, line 349: The default id of this section is "message-consumption" and because the title has the same section id, the links will end up being scoped as "message-consumption-1" and "message-consumption-2". 
To avoid any confusion, I would suggest we either put a section id on this or rename it to disambiguate it. I am in favor of the former option and give it a name of "Consuming records". Wdyt?

onobc (changes requested):
@KartikShrivastava - thank you for this excellent contribution! I have one minor suggestion, but other than that this is good-to-go. 

> [!NOTE]
> Thank you for including the before and after screenshots. Taking the extra time to do this for the ease of your reviewers is one of my favorite things. Much appreciated.

How do you feel about doing this same exact thing to the Reactive section?

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/message-consumption.adoc
KartikShrivastava: Nice catch!
And I liked the idea of renaming it to Consuming records with the similar id. Will make the change

KartikShrivastava:
Thanks for the kind words @onobc!

> How do you feel about doing this same exact thing to the Reactive section?

Totally up for the change, it will also improve the readability. Shall I add it as a separate PR?

onobc:
> Totally up for the change, it will also improve the readability. Shall I add it as a separate PR?

Excellent! And yeh, I think a separate PR is good. Once you update this PR I will merge.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/message-consumption.adoc
KartikShrivastava: Made the changes, please have a look

PR #630, opened 2024-04-05
Introduce consistent transaction settings

onobc:
This commit introduces a common transaction settings class used by `PulsarTemplate` and `PulsarContainerProperties`. This ensures common transaction properties and handling for both components.

See #27

cc: @jonas-grgt

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
@@ -45,6 +45,7 @@ import org.springframework.pulsar.listener.Acknowledgement;
 import org.springframework.pulsar.listener.ConcurrentPulsarMessageListenerContainer;
 import org.springframework.pulsar.listener.PulsarConsumerErrorHandler;
 import org.springframework.pulsar.listener.PulsarContainerProperties;
+import org.springframework.pulsar.listener.PulsarContainerProperties.TransactionSettings;
 import org.springframework.pulsar.listener.PulsarMessageListenerContainer;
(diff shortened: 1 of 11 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/config/MethodPulsarListenerEndpoint.java
```diff
@@ -186,17 +187,24 @@ protected AbstractPulsarMessageToSpringMessageAdapter<V> createMessageListener(
...
-			this.logger.debug(() -> "Listener w/ id [%s] requested no transactions - setting txn mgr to null"
-				.formatted(this.getId()));
-			containerProps.setTransactionManager(null);
+	private void validateAndAdjustTransactionSettings(TransactionSettings txnProps) {
+		// If user did not specify transactional attribute do nothing
+		if (this.transactional == null) {
 			return;
 		}
-		if (containerProps.getTransactionManager() == null) {
+		Assert.state(!(txnProps.isRequired() && !this.transactional),
```
onobc, line 195: These variants need tests.

spring-pulsar/src/main/java/org/springframework/pulsar/listener/PulsarContainerProperties.java
```diff
@@ -335,4 +303,71 @@ private <T> void applyPropIfSpecified(String key, Consumer<T> setter) {
 		}
 	}
 
+	/**
+	 * Transaction related settings.
+	 *
+	 * @since 1.1.0
+	 */
+	public static class TransactionSettings extends TransactionProperties {
```
onobc, line 311: These need sanity tests added - especially the `determineTransactionDefinition` method.

PR #631, opened 2024-04-09
Splits Reactive Support doc into multiple documents

KartikShrivastava:
- This PR is a follow up enhacement of spring pulsar documentation, and splits `reactive-pulsar.adoc` into multiple documents added as sub-section to improve its readability
- Last PR [link](https://github.com/spring-projects/spring-pulsar/pull/627) which did the same to `pulsar.adoc`

### What does it do
- Moves sub-sections in `reactive-pulsar.adoc` to dedicated adocs within newly added `reactive-pulsar` directory
- Fixes anchor references after migration of sub-sections
- Preface is kept in the `reactive-pulsar.adoc`

### Screenshots
Before:
Screenshot from 2024-04-09 22-59-51

After:
Screenshot from 2024-04-09 23-00-02

cc: @onobc


KartikShrivastava:
Hey @onobc, I got a question while making this change:
- I noticed that `Topic Resolution` present in `reactive-pulsar.adoc` is same as that of imperative, so why did we add it as a separate sub-section? I've replicated the similar behaviour, but imho we can directly link the imperative `Topic Resolution` to avoid confusion

onobc:
> Hey @onobc, I got a question while making this change:
> 
> * I noticed that `Topic Resolution` present in `reactive-pulsar.adoc` is same as that of imperative, so why did we add it as a separate sub-section? I've replicated the similar behaviour(by creating a separate section for it), but imho we can directly link the imperative `Topic Resolution` to avoid confusion

Hi @KartikShrivastava ,

So both imperative and reactive docs just included the same content as I could not find a good location for the topic resolution outside of the main sections at the time. Now that they are split into sections though, it may be fine to pull topic-resolution out to its own main section (just above `Pulsar Administration`). If you want to give it a try and see how it looks/reads and if its good then add a commit here for that would be great. 

Either way (even if we leave them inline in each doc) let's move the `topic-resolution.adoc` back to the root level as it looks like we moved it under `pulsar/` in our last refactoring (which is not a problem but it probably makes more sense under the root dir).

Thanks

onobc (changes requested):
@KartikShrivastava nice work! Everything looks good locally. One minor suggestion from me and improvement from you on pulling the topic-resolution up a level. Other than that it is approved.

onobc (approved):
@KartikShrivastava I am going to merge this as-is in order to get it into the RC1. I created a follow up issue https://github.com/spring-projects/spring-pulsar/issues/644 for the refactoring of the topic-resolution.adoc's location. Thanks again for this contribution!

PR #668, opened 2024-05-01
Add docs for transaction support

### LINKED ISSUE #660: Add ref docs for newly added transaction support
Need to add ref docs for the newly added transaction support.

onobc:
Resolves #660

Best way to review...

Pull down locally and follow https://github.com/spring-projects/spring-pulsar/wiki/Working-with-the-Code#updating-the-reference-documentation

### DIFF
--- a/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
+++ b/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
@@ -1,0 +1,222 @@
+[[transactions]]
+= Transactions
+
+This section describes how Spring for Apache Pulsar supports transactions.
+
+[[overview]]
+== Overview
+Spring for Apache Pulsar transaction support is built upon the https://docs.spring.io/spring-framework/reference/data-access/transaction.html[transaction support] provided by Spring Framework.
+At a high-level, transactional resources are registered with a transaction manager which in turn handles the transactional state (commit, rollback, etc..) of the registered resources.
+
+Spring for Apache Pulsar provides the following:
+
+* `PulsarTransactionManager` - used with normal Spring transaction support (`@Transactional`, `TransactionTemplate`, etc)
+* Transactional `PulsarTemplate`
+* Transactional `@PulsarListener`
+* Transaction synchronization with other transaction managers
+
+NOTE: Transaction support has not been added to the Reactive components yet
+
+Transaction support is disabled by default.
+To enable support when using Spring Boot, simply set the `spring.pulsar.transaction.enabled` property.
+Further configuration options are outlined in each component section below.
+
+== Transactional Publishing with `PulsarTemplate`
+All send operations on a transactional `PulsarTemplate` look for an active transaction and enlist each send operation in the transaction (if one is found).
+
+=== Non-transactional use
+By default, a transactional `PulsarTemplate` can also be used for non-transactional operations.
+When an existing transaction is not found it will continue the send operation in a non-transactional fashion.
+However, if the template is configured to require transactions then any attempt to use the template outside the scope of a transaction results in an exception.
+
+TIP: A transaction can be started by a `TransactionTemplate`, a `@Transactional` method, calling `executeInTransaction`, or by a transactional listener container.
+
+=== Local Transactions
+We use the term "local" transaction to denote a Pulsar native transaction that is *not managed* by or associated with Spring's transaction management facility (i.e. `PulsarTransactionManager`).
+Conversely, a "synchronized" transaction is one that *is managed* by or associated with the `PulsarTransactionManager`.
+
+You can use the `PulsarTemplate` to execute a series of operations within a local transaction.
+The following example shows how to do so:
+
+[source, java]
+----
+var results = pulsarTemplate.executeInTransaction((template) -> {
+    var rv = new HashMap<String, MessageId>();
+    rv.put("msg1", template.send(topic, "msg1"));
+    rv.put("msg2", template.send(topic, "msg2"));
+    return rv;
+});
+----
+
+The argument in the callback is the template instance that the `executeInTransaction` method was invoked on.
+All operations on the template are enlisted in the current transaction.
+If the callback exits normally, the transaction is committed.
+If an exception is thrown, the transaction is rolled back.
+
+NOTE: If there is a synchronized transaction in process, it is ignored and a new "nested" transaction is used.
+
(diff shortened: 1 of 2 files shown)


spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
```diff
@@ -0,0 +1,222 @@
...
+To enable support when using Spring Boot, simply set the `spring.pulsar.transaction.enabled` property.
+Further configuration options are outlined in each component section below.
+
+== Transactional Publishing with `PulsarTemplate`
+All send operations on a transactional `PulsarTemplate` look for an active transaction and enlist each send operation in the transaction (if one is found).
+
+=== Non-transactional use
+By default, a transactional `PulsarTemplate` can also be used for non-transactional operations.
+When an existing transaction is not found it will continue the send operation in a non-transactional fashion.
+However, if the template is configured to require transactions then any attempt to use the template outside the scope of a transaction results in an exception.
```
cppwfs, line 30: What exception?

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
```diff
@@ -0,0 +1,222 @@
...
+NOTE: If there is a synchronized transaction in process, it is ignored and a new "nested" transaction is used.
+
+=== Configuration
+The following transaction settings are available directly on the `PulsarTemplate` (via the `transactions` field):
+
+* `enabled` - whether the template supports transactions (default `false`)
+
+* `required` - whether the template requires transactions (default `false`)
+
+* `timeout` - duration of the transaction timeout (default `null`)
```
cppwfs, line 65: Timeout in what?  i.e. seconds/millis/etc

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
```diff
@@ -0,0 +1,222 @@
...
+=== Transactions with Record Listeners
+The above example uses a record listener.
+When using a record listener, a new transaction is created on every listener method invocation which equates to a transaction per message.
+
+NOTE: Because the transaction boundary is per message and each message acknowledgement is enlisted in each transaction, batch ack mode  can not be used with transactional record listeners.
+
+=== Transactions with Batch Listeners
+When using a batch listener, a new transaction is created on every listener method invocation which equates to a transaction per batch of messages.
+
+NOTE: Due to complexity, transactional batch listeners do not currently support custom error handlers.
```
cppwfs, line 121: I don't think we need to say "Due to complexity".   Rather just say, "Transactional batch listeners do not currently support custom error handlers."

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
```diff
@@ -0,0 +1,222 @@
...
+----
+@Bean
+ConcurrentPulsarListenerContainerFactoryCustomizer containerCustomizer() {
+    return (containerFactory) -> containerFactory.getContainerProperties().transactions().setTimeout(Duration.ofSeconds(45));
+}
+----
+
+==== `@PulsarListener`
+By default, each listener respects the transactional settings of its corresponding listener container factory.
+However, the user can set the `transactional` attribute on each `@PulsarListener` to override the container factory setting as follows:
```
cppwfs, line 154: i'm a little confused, it looks like on line 154 our listeners can override the transaction attributes of the factory.   BUt the points below say that we'll get an exception if we disable it.   But I could be misreading it.

cppwfs (commented):
Looks great!  Just a couple of nitpicks

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
onobc: Maybe it would be stated as:
> However, if the template is configured to require transactions and an existing transaction is not found, then it will not continue the send operation and instead will throw an exception.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
onobc: It is an actual `Duration` (i.e. implied in the passed in instance)

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
onobc: The points below are to outline the rules of setting `@PulsarListener(transactional = "")`. 

The only thing we can override really is "enabled" and its conditional on the following rules: 

- If `transactional` not specified - listener inherits settings from container factory (as if the attribute does not exist)

- if `transactional = false` and container factory does not have `required = true` then it will disable txn for listener

- if `transactional = false` and container factory does have `required = true` then txn are always required and this will lead to exception

- if `transactional = true` but container factory does have `enabled = true` then txn are not even available for the container factory, much less for individual listeners (we log this fact if it happens)

I am totally up for suggestions to make this less confusing. It is a bit confusing.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
cppwfs: Just curious if we could say what type of exception is being thrown.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
cppwfs: After rereading the rules again it makes sense.    I was getting required and enabled confused.

spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/transactions.adoc
onobc: I try not to do that in the docs because that can get out of sync easy because it is implementation specific (e.g. IllegalStateException today but maybe IllegalArgumentException next month).

PR #678, opened 2024-05-17
Fix payload type in Reactive tombstone doc

### LINKED ISSUE #667: Copy/paste artifact in the Consuming Null Payloads docs
See more info in the https://docs.spring.io/spring-pulsar/docs/1.1.0-RC1/reference/reference/tombstones-reactive.html#tombstones-reactive.consume.

The table talks about `Flux<org.apache.pulsar.client.api.Messages<T>>` where it feels like this has to be Flux<org.springframework.messaging.Message<T>>.
Although, that indeed rises the question how about `org.apache.pulsar.client.api.Messages` in reactive manner...

onobc:
Also clarify that `org.apache.pulsar.client.api.Messages<T>` are not supported in `@ReactivePulsarListener`.

Fixes #667


onobc:
PTAL @artembilan

artembilan (approved):
LGTM.
I don't have permissions to merge.

PR #819, opened 2024-08-29
Fix invalid configuration property references

vpavic:
This commit fixes invalid documentation references to `spring.pulsar.consumer.subscription.type` configuration property.

### DIFF
--- a/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/message-consumption.adoc
+++ b/spring-pulsar-docs/src/main/antora/modules/ROOT/pages/reference/pulsar/message-consumption.adoc
@@ -15,7 +15,7 @@ Spring Boot provides this consumer factory which you can further configure by sp
 
 TIP: The `spring.pulsar.consumer.subscription.name` property is ignored and is instead generated when not specified on the annotation.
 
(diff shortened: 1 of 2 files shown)


onobc (approved):
LGTM. Thanks for the fix @vpavic

vpavic:
Thanks for merging - though it's funny how commit subject line ended up being something completely different/unrelated when merged on `main`. :slightly_smiling_face:

I suggest you consider amending the commit and force pushing before further commits land on `main`.

onobc:
>  Thanks for merging - though it's funny how commit subject line ended up being something completely different/unrelated when merged on main. 🙂

Yeh, I did the rebase/merge in Github (which I rarely do) and I did not notice the wonky commit message. Seems like I am trying to self-promote lol - but trust me! I am not :) 

> I suggest you consider amending the commit and force pushing before further commits land on main.

Believe me, it crossed my mind and I was very tempted to do so. We have a hard policy never to force push on main. Personally, I am not against it in rare cases (like this). It would not hurt anything and it will avoid a ridiculous commit message. But I left it alone. 

Lesson here is.... slow down and re-read your commits before pushing the button Chris. :)

Issue #816, opened 2024-08-23
PulsarListener does not recover after it has startup failure


eljefe6a:
I'm hitting an issue where the consumption randomly stops working. I haven't seen any pattern to it. I'm using spring-pulsar 1.1.0.

I've spent the past 4 hours trying to figure out why data wasn't being saved to the database. It turned out that the spring-pulsar consumers weren't receiving data. I added more logging to the `@PulsarListener` annotated method to log out the object after the first line of code. It still didn't log out. I added more logging to the `@PulsarListener` annotated method, and it started working. I restarted the container (with no code changes), and it stopped logging. I restarted again with no code changes, and it started logging again.

The methods are consuming an Avro topic. The data is being produced, and I verified with the `pulsar-admin` that new messages are arriving. There are five different methods in that class that are `@PulsarListener` listener annotated. With this bug, some methods will receive messages and others won't. No exceptions are being logged, except some initial exclusive consumer exceptions while the other container is being terminated that stop being logged after startup. The methods are using both `topicPattern` and `topics`.  The process is running in a K8s container.

Is there any known limitation or bug that I'm hitting? It makes no sense that the methods stop receiving data with no code changes.

onobc:
Hi @eljefe6a,

> There are five different methods in that class that are `@PulsarListener` listener annotated

The 5 listeners consuming from the same topic,  are you wanting each listener (method) to consume every message or have them shared across the listeners?

It sounds like all methods are using the same subscription in an exclusive mode - which won't work and could explain the random behavior.

Are you specifying `subscriptionType` and/or `subscriptionName` on the `@PulsarListener` annotation? If not specified, the default type is exclusive.

Anytime I start mentioning subscriptions, I always do myself a favor and look at the nice diagram from the Pulsar docs that [details the flexible subscription model](https://pulsar.apache.org/docs/3.3.x/concepts-messaging/#subscriptions).

Can you share the 5 methods (just signatures and listener annotations)?

Thanks

eljefe6a:
They are consuming five different topics. Each one is using a different subscriber name.

Here two examples:
```
@PulsarListener(subscriptionName = "${spring.pulsar.client.configurations.subscription-prefix}-OrderInfoWebserviceListener-sub", topicPattern = "${spring.pulsar.client.configurations.league-topic-pattern}", schemaType = SchemaType.AVRO)
```

```
[...]

Issue #445, opened 2023-09-13
Add ability for user to configure whether to fail fast on startup


nerdBoi:
## Application fails to connect to Pulsar; however, it swallows / logs the exception and does not restart

Is there anyway to add a configuration so that if a failure to connect to Pulsar happens on boot, it restarts?
For context this is what happened:
 We applied a rolling restart to all of our applications
 Spring Boot Pulsar Client tried to connect to pulsar; however, pulsar was not finished spinning up. 
 So, Pulsar just gave us back connection refused.

Error was like this:
```
java.util.concurrent.CompletionException: org.apache.pulsar.shade.io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: \"},{\"attempt\":5,\"error\":\"java.util.concurrent.CompletionException: 
org.apache.pulsar.client.api.PulsarClientException: java.util.concurrent.CompletionException: org.apache.pulsar.shade.io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: \"},{\"attempt\":6,\"error\":\"java.util.concurrent.CompletionException: 
org.apache.pulsar.client.api.PulsarClientException: java.util.concurrent.CompletionException: org.apache.pulsar.shade.io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: \"}]}\n\tat org.apache.pulsar.client.api.PulsarClientException.unwrap(PulsarClientException.java:1074)\n\tat org.apache.pulsar.client.impl.ConsumerBuilderImpl.subscribe(ConsumerBuilderImpl.java:104)\n\tat 
org.springframework.pulsar.core.DefaultPulsarConsumerFactory.createConsumer(DefaultPulsarConsumerFactory.java:89)\n\tat org.springframework.pulsar.listener.DefaultPulsarMessageListenerContainer$Listener.<init>(DefaultPulsarMessageListenerContainer.java:298)\n\tat org.springframework.pulsar.listener.DefaultPulsarMessageListenerContainer.doStart(DefaultPulsarMessageListenerContainer.java:126)\n\tat org.springframework.pulsar.listener.AbstractPulsarMessageListenerContainer.start(AbstractPulsarMessageListenerContainer.java:110)\n\tat 
```

Final note, this is my first time submitting an issue on an open source project. If I goofed up just let me know please.

onobc:
Hi @nerdBoi ,
[...]

PR #823, opened 2024-09-01
Add startup failure policy to listeners

onobc:
Previously, when a listener container failed to start, it would only log the exception. This commit introduces `StartupFailurePolicy` that allows listener containers to CONTINUE, STOP, RETRY when an error is encountered on startup.

See #445
See #816


onobc:
Closing in favor of #824

Issue #820, opened 2024-08-30
Listener concurrency attribute is inconsistent


onobc:
The `concurrency` attribute for the non-reactive listener should be specified on the `PulsarContainerProperties` rather than the `ConcurrentPulsarMessageListenerContainerFactory`.

The `ConcurrentPulsarListenerContainerFactory#initializeContainer` method can simply ask the `PulsarContainerProperties` for the concurrency rather than store it.

### Motivation

- consistency w/ Reactive counterpart
- consistency w/ other container properties
- simplify SpringBoot property mapping ([example](https://github.com/spring-projects/spring-boot/pull/42062/files))

vpavic:
As noted in https://github.com/spring-projects/spring-boot/pull/42062 I can pick this one up so feel free to assign it to me. I'll have some time tomorrow to submit a PR.

onobc:
Thanks for another great contribution @vpavic !

PR #826, opened 2024-09-03
Add concurrency to PulsarContainerProperties

vpavic:
This commit adds concurrency property to `PulsarContainerProperties` instead of managing it directly on
`ConcurrentPulsarListenerContainerFactory`, which provides consistency with both reactive counterpart and container properties in general.

Resolves: gh-820


onobc (approved):
LGTM - thanks for the improvement @vpavic

Issue #998, opened 2025-01-14
Listener consumes the same message after it is sent to the DLT topic


darshimo:
# Problem

I am using Spring Boot 3.4.1 and Spring Pulsar 1.2.1.

A listener will consume the same message indefinitely if the following conditions are met:

- The listener is a batch listener with `PulsarConsumerErrorHandler`.
- The topic contains only one message.
- The listener throws a `PulsarBatchListenerFailedException` on the message.

Below is my code. The expected behavior is “after a message is sent to the DLT topic, the listener will no longer receive that message”.

```java
@Bean
ApplicationRunner runner(final PulsarTemplate<String> client) {
  return (args) -> client.send("my-topic", "invalid");
}

@PulsarListener(
    subscriptionName = "my-sub",
    topics = "my-topic",
    pulsarConsumerErrorHandler = "pulsarConsumerErrorHandler",
    batch = true,
    ackMode = AckMode.MANUAL
)
void listen(final List<Message<String>> messages, Acknowledgement ack) {
  for (Message<String> message : messages) {
    if ("invalid".equals(message.getValue())) {
      System.out.println("Message Refused: " + message.getValue());
      throw new PulsarBatchListenerFailedException("failed", message);
    }
    System.out.println("Message Received: " + message.getValue());
    ack.acknowledge(message.getMessageId());
  }
}

@PulsarListener(topics = "my-dlt")
void listenDlt(String msg) {
  System.out.println("From DLT: " + msg);
}

@Bean
PulsarConsumerErrorHandler<String> pulsarConsumerErrorHandler(PulsarTemplate<String> pulsarTemplate) {
  return new DefaultPulsarConsumerErrorHandler<>(
      new PulsarDeadLetterPublishingRecoverer<>(
          pulsarTemplate,
          (c, m) -> "my-dlt"
      ),
      new FixedBackOff(100, 2)
  );
}
```

I got the following log:

```
Message Refused: invalid
Message Refused: invalid
Message Refused: invalid
From DLT: invalid
Message Refused: invalid
Message Refused: invalid
Message Refused: invalid
From DLT: invalid
Message Refused: invalid
Message Refused: invalid
Message Refused: invalid
From DLT: invalid
...
```

# Analysis

The following section of the source code seems to be the cause.

[...]

PR #1002, opened 2025-01-16
Fix batch listener error handling

darshimo:
### Current behavior

A listener continuously consumes the same message indefinitely under the following conditions:

- The listener is a batch listener with `PulsarConsumerErrorHandler`.
- The listener receives a batch that contains only one message.
- The listener throws a `PulsarBatchListenerFailedException` for that message.

### Change

After handling an error, the batch—excluding the head message—is passed back to the listener, regardless of the original batch length.

### Reference issue

 #998

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/listener/DefaultPulsarMessageListenerContainer.java
@@ -781,15 +781,8 @@ public class DefaultPulsarMessageListenerContainer<T> extends AbstractPulsarMess
 				this.pulsarConsumerErrorHandler.recoverMessage(this.consumer, pulsarMessage,
 						pulsarBatchListenerFailedException);
 				handleAck(pulsarMessage, txn);
-				if (messageList.size() == 1) {
-					messagesPendingInBatch.set(false);
-				}
-				else {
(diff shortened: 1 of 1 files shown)


onobc:
@darshimo can you sign the commit following [the DCO process](https://github.com/spring-projects/spring-pulsar/blob/main/CONTRIBUTING.adoc#developer-certificate-of-origin)? Thanks.

darshimo:
@onobc
Thank you for your feedback. I've added the DCO sign-off to the commit. Please let me know if there is anything else I should do. Thanks.

onobc:
@darshimo thanks again for the contribution - nice work!. 

I added a subsequent [commit](https://github.com/spring-projects/spring-pulsar/commit/16e0c9ee8c2bafa6a9b16f9dc4fb977eb2b71545) that adds a test and only uses `List.remove` in the special case (`sublist` performs better than `remove`).

I would **normally** have added these 2 items during the PR review **but**....

**1.** I wanted to get this merged for the upcoming release this Tuesday and I just now got time to do the review so another code review round would not have been possible

**2.** For the added test case, the batch error tests need some refactoring as they highly violate the DRY principle and it is a bit confusing as to what the expected verify counts are even for. I did not want to put this burden on you. That being said, I am going to refactor these tests in a subsequent proposal to reduce duplication and make it clear as to what they are doing.

Closing in favor of #fddc711730b8891131aa2d9a6975b02f0fdf23c8

PR #1004, opened 2025-01-18
Fix batch listener error handling [VERIFY]

onobc:


onobc:
Was only using to double-check polish commit changes in CI

PR #1064, opened 2025-03-17
Remove no longer needed exclusions from pulsar-client-all

### LINKED ISSUE #963: Remove no longer needed exclusions from pulsar-client-all
Pulsar cleaned up quite a few dependencies in `pulsar-client-all` in https://github.com/apache/pulsar/pull/23647.

We have most of these libs excluded in [spring-pulsar](https://github.com/spring-projects/spring-pulsar/blob/475c515ec53aa20a4d50069fffa2c6dd9797fdc9/spring-pulsar/spring-pulsar.gradle#L10-L15) but can now safely remove these exclusions.

therepanic:
The improvements introduced in [this PR](https://github.com/apache/pulsar/pull/23647) include handling `javax.activation` via shading and migrating from javax to jakarta. Given these changes, we should remove some of the now unnecessary exclusions.

Fix #963

### DIFF
--- a/spring-pulsar/spring-pulsar.gradle
+++ b/spring-pulsar/spring-pulsar.gradle
@@ -9,7 +9,6 @@ dependencies {
 	api 'io.micrometer:micrometer-observation'
 	api (libs.pulsar.client.all) {
 		exclude group: 'org.apache.logging.log4j'
-		exclude group: 'com.sun.activation', module: 'javax.activation'
 		exclude group: 'javax.validation', module: 'validation-api'
 		exclude group: 'com.google.protobuf', module: 'protobuf-java'
 	}


onobc:
Hi @panic08 - great to see you over here in this project 👋🏻 

We just released `1.1.10` and `1.2.4` but will get this merged in soon. We intentionally did not add this change in yet as shadowing, transitive deps, and dependencies in general can be risky beyond the surface, especially when we have something like:

- spring-boot (uses)
   - pulsar-client-all
   - spring-pulsar
   - pulsar-client-reactive
**and**  
- spring-pulsar (uses)
   - pulsar-client-all
   - pulsar-client-reactive
**and**
- pulsar-client-reactive (uses)
    - pulsar-client-all

Thank you very much for the contribution. Once we branch main into `1.2.x` and main is then our target for `2.0.0` we will merge this.

Issue #1029, opened 2025-02-11
Automate protobuf generated test POJOs


onobc:
The Protobuf POJOs used for testing recently got out of sync w/ reality and tests started failing due to:

>java.lang.UnsupportedOperationException: As of 2022/09/29 (release 21.7) makeExtensionsImmutable should not be called from protobuf gencode. If you are seeing this message, your gencode is vulnerable to a denial of service attack. You should regenerate your code using protobuf 25.6 or later. Use the latest version that meets your needs. However, if you understand the risks and wish to continue with vulnerable gencode, you can set the system property `-Dcom.google.protobuf.use_unsafe_pre22_gencode` on the command line. See security vulnerability: https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-h4h5-3hr4-j3g2

Just needed to update the generated POJOs. However, we should automate the process to regeenerate the files when building using a plugin such as https://github.com/google/protobuf-gradle-plugin

onobc:
Fixed via https://github.com/spring-projects/spring-pulsar/pull/1068/commits/59f6b02628b19afc8955c33d59347ccbfbbb9779

PR #1068, opened 2025-03-20
Automate protobuf generation in tests

panic08:
I think it's not a good idea to generate a class from a protofile manually and leave an instruction for it, to avoid any future problems as it happened with #1029, and it's best to automate it and do it in the build.

You may notice that in the modules I disabled the `extractIncludeProto` and so on, as they interfered with the generation of the protoclass. Also spring checkformat was breaking the assembly, so I disabled checkformat for protofiles.

Fix: #1029


onobc:
Another excellent contribution! Thanks @panic08 .

Closing in favor of https://github.com/spring-projects/spring-pulsar/commit/ed56a3941cc79b42e1798a4de7d8ed398385aad2. 

I also added a small polish to it [here](https://github.com/spring-projects/spring-pulsar/commit/1a4ee049a0380ae28818f3aa8375c38b2ea79410#r154593970)

PR #1104, opened 2025-05-01
Fix listener TaskExecutor config

### LINKED ISSUE #1103: Unable to configure virtual thread based Executor
Spring Pulsar version: `1.2.4`
Spring Boot version: `3.4.4`
JDK version is `21`

I tried to switch the threads on which the `@PulsarListener` methods run to virtual threads, but the thread name remained prefixed with `EndpointContainer#0-0-C`.

I'm unsure if I'm missing something, but I suspect the virtual thread setup doesn't work.

## Why it doesn't seem to work

Firstly, because the `@PulsarListener` annotated method doesn't log using the custom thread prefix I set.
If the Spring Boot level virtual thread usage is configured, it doesn't use that thread prefix either.

Secondly, I added the following code snippet to a `@SpringBootTest`:

```java
final List<String> containerClassTypes = applicationContext
    .getBean(PulsarListenerContainerRegistry.class)
    .getAllListenerContainers()
    .stream()
    .map(container -> {
        if (container instanceof ConcurrentPulsarMessageListenerContainer listenerContainer) {
            return listenerContainer.getContainerProperties().getConsumerTaskExecutor().getClass().getSimpleName();
        } else {
            return "Not concurrent pulsar listener container.";
        }
    })
    .toList();
```

The list only contained `SimpleAsyncTaskExecutor` values.

### Attempt 1

I set the `spring.threads.virtual.enabled` property to `true`.

### Attempt 2

I created a container factory customizer bean, as described in the documentation.

```java
@Configuration
public class PulsarVirtualThreadConfig {

    @Bean
    public PulsarContainerFactoryCustomizer<ConcurrentPulsarListenerContainerFactory<Object>> customPulsarVirtualThreadListenerCustomization() {
        return factory -> factory
            .getContainerProperties()
            .setConsumerTaskExecutor(new VirtualThreadTaskExecutor("pulsar-consumer-virtual-thread-"));
    }

}
```

## Possible reason for the problem

After debugging the Spring Pulsar code for a while, it seems to me that at some point there is a manual deep clone of container properties, and the executor property isn't copied. Because of this it'll always be `null`, and will always be set to a `SimpleAsyncTaskExecutor`.

### Debugging steps in the `v1.2.5` codebase

1. `PulsarListenerEndpointRegistrar:82`
   1. It seems the `@PulsarListener` registrations start here, called by Spring during initialization.
2. `PulsarListenerEndpointRegistrar:91`
   1. The factory here contains the `VirtualThreadTaskExecutor` set using either method described earlier.
3. `ConcurrentPulsarListenerContainerFactory:122`
   1. This is the last place where we can observe the factory properties containing the `VirtualThreadTaskExecutor` in the `consumerTaskExecutor` field.
   2. The `PulsarContainerProperties` object passed into the new instance of `ConcurrentPulsarMessageListenerContainer` does not have its `consumerTaskExecutor` field set.
4. When the `doStart` method is called on `ConcurrentPulsarMessageListenerContainer`, it'll eventually call the `configureChildContainer` method. Here, at line 124, the `consumerTaskExecutor` is always `null`, and a new `SimpleAsyncTaskExecutor` will be created.

garfailed:
Fixes #1103

I noticed there are missing unit tests for the listener container default property setup. I tried creating them, but it seems like a larger piece of work with all the factory / endpoint / default branches.

Should I work on that and create a separate PR for it, or would it not add that much value?

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java
@@ -37,6 +37,7 @@ import org.springframework.util.StringUtils;
  * @author Chris Bono
  * @author Alexander Preuß
  * @author Vedran Pavic
+ * @author Daniel Szabo
  */
 public class ConcurrentPulsarListenerContainerFactory<T>
 		extends AbstractPulsarListenerContainerFactory<ConcurrentPulsarMessageListenerContainer<T>, T> {
@@ -80,6 +81,7 @@ public class ConcurrentPulsarListenerContainerFactory<T>
 		var containerProps = new PulsarContainerProperties();
 
 		// Map factory props (defaults) to the container props
+		containerProps.setConsumerTaskExecutor(factoryProps.getConsumerTaskExecutor());
 		containerProps.setSchemaResolver(factoryProps.getSchemaResolver());
 		containerProps.setTopicResolver(factoryProps.getTopicResolver());
 		containerProps.setSubscriptionType(factoryProps.getSubscriptionType());
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/config/AbstractPulsarListenerContainerFactory.java
@@ -115,7 +115,6 @@ public abstract class AbstractPulsarListenerContainerFactory<C extends AbstractP
 
 		endpoint.setupListenerContainer(instance, this.messageConverter);
 		initializeContainer(instance, endpoint);
-		// customizeContainer(instance);
 		return instance;
 	}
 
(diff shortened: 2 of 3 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java
```diff
@@ -80,6 +81,7 @@ protected ConcurrentPulsarMessageListenerContainer<T> createContainerInstance(Pu
 		var containerProps = new PulsarContainerProperties();
 
 		// Map factory props (defaults) to the container props
+		containerProps.setConsumerTaskExecutor(factoryProps.getConsumerTaskExecutor());
```
garfailed, line 84: I didn't do the null check mentioned in the issue comment because this will always be a new instance. Most other options also assume we just want to set the factory values.

onobc:
Thanks for the contribution @garfailed . We will get to the review here shortly. 

> I noticed there are missing unit tests for the listener container default property setup. I tried creating them, but it seems like a larger piece of work with all the factory / endpoint / default branches.
> Should I work on that and create a separate PR for it, or would it not add that much value?

Adding the tests would likely be a valuable thing, especially if it would have revealed this issue up front. Since you have already added a test to catch this fix, I would add those other tests in a separate PR.

PR #1177, opened 2025-07-16
Stop shading checker-qual in caffeine cache provider (1.2.x line)

onobc:
This commit adjusts the shading in the spring-pulsar-cache-provider-caffeine module to not include the `org.checkerframework:checker-qual` in the shaded jar.


onobc:
@ mbhave PTAL

Issue #1241, opened 2025-09-01
Allow `DeadLetterPolicy.deadLetterTopic` to be non-fully-qualified


onobc:
## Current Behavior
Currently, the `DeadLetterPolicy.deadLetterTopic` must be fully-qualified or it will not be able to find the topic to send the dead message to.

> [!NOTE]
> It may currently allow a relative topic when the `defaults.topic.[tenant|namespace]` are not specified in `application.yml` but need to verify this. 

## Desired Behavior
When the `DeadLetterPolicy.deadLetterTopic` is not fully-qualified , it should automatically use the project’s default `tenant` and `namespace` configuration.  This way, we wouldn’t need to manually write the full path in every `DeadLetterPolicy`, making the configuration more intuitive and less error-prone.

> [!NOTE]
> Discovered in [1216](https://github.com/spring-projects/spring-pulsar/issues/1216#issuecomment-3241139340)

therepanic:
Hi, @onobc! It doesn't look too complicated, I would offer my help if you don't mind.

oneby-wang:
Hi, @onobc Chris, if this fix [feat: Allow DeadLetterPolicy.deadLetterTopic to be non-fully-qualified #1241](https://github.com/oneby-wang/spring-pulsar/commit/baa22fa87eead4bed6cd6ca4aa7e09eedfe31d7a) is ok, I'll create a pull request.

PR #1245, opened 2025-09-05
Allow `DeadLetterPolicy#deadLetterTopic` to be non-fully-qualified

therepanic:
I chose `PulsarListenerAnnotationBeanPostProcessor` because I think it's easier to inject `PulsarTopicBuilder` into it, which I think is better than passing it to `MethodPulsarListenerEndpoint` later.

Closes: gh-1241

### DIFF
--- a/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/DefaultReactivePulsarConsumerFactory.java
+++ b/spring-pulsar-reactive/src/main/java/org/springframework/pulsar/reactive/core/DefaultReactivePulsarConsumerFactory.java
@@ -98,6 +98,18 @@ public class DefaultReactivePulsarConsumerFactory<T> implements ReactivePulsarCo
 			var fullyQualifiedTopics = topics.stream().map(this.topicBuilder::getFullyQualifiedNameForTopic).toList();
 			mutableSpec.setTopicNames(fullyQualifiedTopics);
 		}
+		if (mutableSpec.getDeadLetterPolicy() != null) {
+			var dlt = mutableSpec.getDeadLetterPolicy().getDeadLetterTopic();
+			if (dlt != null) {
+				mutableSpec.getDeadLetterPolicy()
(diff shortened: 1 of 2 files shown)


spring-pulsar/src/main/java/org/springframework/pulsar/annotation/PulsarListenerAnnotationBeanPostProcessor.java
```diff
@@ -110,6 +113,12 @@ public class PulsarListenerAnnotationBeanPostProcessor<V> extends AbstractPulsar
 
 	private final List<MethodPulsarListenerEndpoint<?>> processedEndpoints = new ArrayList<>();
 
+	@Nullable private ObjectProvider<PulsarTopicBuilder> topicBuilder;
+
+	public PulsarListenerAnnotationBeanPostProcessor(@Nullable ObjectProvider<PulsarTopicBuilder> topicBuilder) {
+		this.topicBuilder = topicBuilder;
+	}
```
therepanic, line 120: I believe this will work? When we specify the tenant and namespace in application.properties, we create a bean with an instance of `PulsarTopicBuilder`?

onobc: Closing per recent findings in https://github.com/spring-projects/spring-pulsar/pull/1245#issuecomment-3293566884

therepanic:
Hi, @onobc!

Sorry for the late reply. Yes, your suggestion is actually much simpler than what I tried to propose. I did what you suggested, also for the reactive version of `DefaultPulsarConsumerFactory`.

onobc:
Thanks for another great contribution @therepanic !

I added a [commit](https://github.com/spring-projects/spring-pulsar/commit/c5cc1f91a75969eb5734f12762c95a78c3dad381)  to add tests for the dead/retry topics as well.

Closing in favor of https://github.com/spring-projects/spring-pulsar/commit/64a9262ac2dd2ff050c8927ab490728c8f4ba8cd

PR #1506, opened 2026-06-05
Propagate startup failure properties to created container

### LINKED ISSUE #1505: ConcurrentPulsarListenerContainerFactory customizations not respected
Ran into this trying to set retry policy. This should retry indefinitely, but it fails using the default retry policy instead of the customized one.

> `Configured to stop on startup failures - exiting`

```
import org.junit.jupiter.api.Test;
import org.springframework.boot.SpringBootConfiguration;
import org.springframework.boot.autoconfigure.ImportAutoConfiguration;
import org.springframework.boot.pulsar.autoconfigure.PulsarAutoConfiguration;
import org.springframework.boot.pulsar.autoconfigure.PulsarContainerFactoryCustomizer;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.core.retry.RetryPolicy;
import org.springframework.core.retry.RetryTemplate;
import org.springframework.pulsar.annotation.PulsarListener;
import org.springframework.pulsar.config.ConcurrentPulsarListenerContainerFactory;
import org.springframework.pulsar.config.StartupFailurePolicy;

import java.time.Duration;

import static org.junit.jupiter.api.Assertions.assertTrue;

@SpringBootTest(classes = RetryTest.Config.class)
public class RetryTest {
    @Configuration
    @SpringBootConfiguration
    @ImportAutoConfiguration({PulsarAutoConfiguration.class})
    static class Config {
        @Bean
        public PulsarContainerFactoryCustomizer<ConcurrentPulsarListenerContainerFactory<?>> containerFactoryCustomizer() {
            return (factory) -> {
                var retryPolicy = RetryPolicy.builder()
                        .maxRetries(Long.MAX_VALUE)
                        .delay(Duration.ofSeconds(30))
                        .build();
                factory.getContainerProperties().setStartupFailureRetryTemplate(new RetryTemplate(retryPolicy));
                factory.getContainerProperties().setStartupFailurePolicy(StartupFailurePolicy.RETRY);
            };
        }
        @PulsarListener(topics = "topic", subscriptionName = "topic-sub")
        void consume(String msg) {
            // noop
        }
    }

    @Test
    public void retryTest() {
        assertTrue(false);
    }
}
```

[Looking at the code](https://github.com/spring-projects/spring-pulsar/blob/main/spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java#L70) there are several  container properties that are not passed through and the child container doesn't read them from the factory either. When I debugged this the `ConcurrentPulsarMessageListenerContainer#getContainerProperties` call returned the customized properties. Ironic considering that this approach is given as an example of how to set `StartupFailurePolicy.RETRY`.

seonwooj0810:
Fixes #1505

## Problem

`ConcurrentPulsarListenerContainerFactory#createContainerInstance` creates a fresh `PulsarContainerProperties` for every container and copies only a curated subset of the factory's container properties onto it (consumer task executor, schema/topic resolvers, subscription type/name, transactions). The **startup failure policy** and **startup failure retry template** are not among them.

As a result, when a user configures `StartupFailurePolicy.RETRY` and/or a custom `startupFailureRetryTemplate` on the factory's container properties — e.g. via a `PulsarContainerFactoryCustomizer`, which is the documented way to opt into startup retries — those settings are silently dropped and the created container keeps the default `STOP` policy (`Configured to stop on startup failures - exiting`).

## Fix

Copy the startup failure retry template and policy from the factory props to the container props in `createContainerInstance`. The retry template is set **before** the policy because `PulsarContainerProperties#setStartupFailureRetryTemplate` implicitly switches the policy to `RETRY` when given a non-null template; the policy is then set explicitly to mirror the factory exactly.

```java
containerProps.setStartupFailureRetryTemplate(factoryProps.getStartupFailureRetryTemplate());
containerProps.setStartupFailurePolicy(factoryProps.getStartupFailurePolicy());
```

## Tests

Added unit tests to `ConcurrentPulsarListenerContainerFactoryTests`:
- `StartupFailurePolicyFrom` — factory's policy is propagated to the created container, and the default (`STOP`) is used when unset.
- `StartupFailureRetryTemplateFrom` — factory's retry template is propagated (and the policy is switched to `RETRY` accordingly).

## Verification done

- `./gradlew :spring-pulsar:test --tests "*ConcurrentPulsarListenerContainerFactoryTests"` — BUILD SUCCESSFUL (new tests pass; compiled with `-Werror` and the project's nullability/errorprone checks).
- `./gradlew :spring-pulsar:checkFormat :spring-pulsar:checkstyleMain :spring-pulsar:checkstyleTest` — BUILD SUCCESSFUL.
- Change is limited to the two files above (a stray pre-existing formatting diff the `format` task surfaced in an unrelated class was reverted).

Note: this PR addresses the startup-failure properties called out by the reproducer; if maintainers prefer, the same propagation block could later be reviewed for any other factory-level properties not yet copied.

### DIFF
--- a/spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java
+++ b/spring-pulsar/src/main/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactory.java
@@ -77,6 +77,10 @@ public class ConcurrentPulsarListenerContainerFactory<T>
 		containerProps.setTopicResolver(factoryProps.getTopicResolver());
 		containerProps.setSubscriptionType(factoryProps.getSubscriptionType());
 		containerProps.setSubscriptionName(factoryProps.getSubscriptionName());
+		// Set the retry template before the policy as setting a non-null template
+		// implicitly switches the policy to RETRY
+		containerProps.setStartupFailureRetryTemplate(factoryProps.getStartupFailureRetryTemplate());
+		containerProps.setStartupFailurePolicy(factoryProps.getStartupFailurePolicy());
 		var factoryTxnProps = factoryProps.transactions();
 		var containerTxnProps = containerProps.transactions();
 		containerTxnProps.setEnabled(factoryTxnProps.isEnabled());
--- a/spring-pulsar/src/test/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactoryTests.java
+++ b/spring-pulsar/src/test/java/org/springframework/pulsar/config/ConcurrentPulsarListenerContainerFactoryTests.java
@@ -24,6 +24,8 @@ import org.apache.pulsar.client.api.SubscriptionType;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
+import org.springframework.core.retry.RetryPolicy;
+import org.springframework.core.retry.RetryTemplate;
 import org.springframework.core.task.AsyncTaskExecutor;
 import org.springframework.pulsar.core.PulsarConsumerFactory;
 import org.springframework.pulsar.listener.PulsarContainerProperties;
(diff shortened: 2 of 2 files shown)


seonwooj0810:
Gentle ping — open since early June.

`ConcurrentPulsarListenerContainerFactory#createContainerInstance` copies only a curated subset of the factory's container properties onto the per-container instance, and the startup failure policy / startup failure retry template are not in that subset — so `StartupFailurePolicy.RETRY` configured via a `PulsarContainerFactoryCustomizer` is silently dropped (#1505).

The PR adds those two to the copied set. Would you rather keep curating individual properties, or is a broader copy the direction you'd prefer? Happy to rework it. Still merges cleanly, checks green.

seonwooj0810:
Following up: without this, `StartupFailurePolicy.RETRY` and a custom `startupFailureRetryTemplate` configured through `PulsarContainerFactoryCustomizer` (the documented way to opt into startup retries) are silently dropped for every container the factory creates (#1505). Branch is clean and CI is green; happy to adjust if needed.

PR #1515, opened 2026-06-09
Fix thread safety of thread-bound transactions in PulsarTemplate

Dream95:
PulsarTemplate stored per-thread transactions in a HashMap keyed by Thread. HashMap is not thread-safe, so concurrent use of a shared PulsarTemplate instance could lose transaction bindings. 

Replace with ThreadLocal to correctly scope transaction state per thread without shared mutable map access.


sobychacko:
Fixed via https://github.com/spring-projects/spring-pulsar/commit/c2c7d4d530b34baa5bb13774a5ffaa6b10a483d1 and https://github.com/spring-projects/spring-pulsar/commit/1862a055a07e2af95441b8bf2d247cf9467e7ef4.

PR #1537, opened 2026-08-04
Align with commercial build support

sobychacko:
Align spring-pulsar's Gradle and CI setup with spring-pulsar-commercial so
the same workflow files and build scripts can run, largely unmodified, in
both repositories.

- Add an `isCI` Gradle variable in a buildscript block and conditionally
  apply a new `commercial-settings.gradle` when the `COMMERCIAL` environment
  variable is set (i.e. only in the commercial repository). Inert in OSS.
- `commercial-settings.gradle` resolves artifacts from the enterprise
  Artifactory repositories in CI, falling back to the spring-commercial
  remote for local development.
- Recognize `*.x-internal` maintenance branches alongside `main` and `*.x`
  in the CI-triggering workflows (ci.yml, ci-pr.yml).
- Relax the repository guard from `github.repository == spring-projects/
  spring-pulsar` to `github.repository_owner == spring-projects` so CI also
  runs in the commercial repository.
- Introduce a `vars.COMMERCIAL` repository variable as the single source of
  truth for repo-specific behavior: it selects the Artifactory URL/repository
  for snapshot deployment and, together with `COMMERCIAL_ARTIFACTORY_*`
  secret fallbacks, the deploy credentials.
- Gate the OSS-only `deploy_docs_antora` and Maven Central `perform_release`
  jobs on `!vars.COMMERCIAL`; commercial docs and releases are handled by the
  release-train workflows.
- Add release-train automation workflows (release-train-build/join/leave/
  ready/retry/test) and their composite actions, plus a `post-release.yml`
  gated on `vars.COMMERCIAL`, and the `.github/workflow-generator.yml` config.

Signed-off-by: Soby Chacko <soby.chacko@broadcom.com>


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,23 @@
...
+on:
+  push:
+    tags:
+      - v*
+
+run-name: Post-release for ${{ github.ref_name }}
+
+jobs:
+  release:
+    # Commercial-only: in OSS, post-release automation runs inside ci.yml's perform_release job.
```
artembilan: I don't think this is correct.
The `ci.yml` is not use for release anymore.
The tag is going to be pushed to repository by the Orchestrator.
Therefore, `on:  push:  tags:` is still valid, and this condition is wrong.

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,23 @@
...
+    if: ${{ vars.COMMERCIAL }}
+    permissions:
+      contents: write
+      issues: write
+
+    uses: spring-io/spring-github-workflows/.github/workflows/spring-post-release.yml@main
```
artembilan, line 16: We need to double check if the logic in that `spring-post-release.yml` is valid for this project: https://github.com/spring-io/spring-github-workflows/blob/main/.github/workflows/spring-post-release.yml

.github/workflows/release-train-test.yml
```diff
@@ -0,0 +1,83 @@
...
+        if [ "$GITHUB_RUN_ATTEMPT" -gt 1 ]; then
+          echo "Re-runs are prohibited. Use the 'Release Train – Retry' workflow to retry test failures"
+          exit 1
+        fi
+    - name: "Set up Java"
+      id: "set-up-java"
+      uses: "actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95" # v5.6.0
+      with:
+        distribution: "liberica"
+        java-version: "17"
```
artembilan: I think the copy/paste artifact.
According to the generator has to be Java 25.
I see the same problem in SA, SK & SI.
Will fix there shortly.

Thank you for doing this, so we spot problems on review! 😄

build.gradle
```diff
@@ -14,6 +18,13 @@ apply from: 'gradle/version-catalog-update.gradle'
 apply from: 'gradle/java-conventions.gradle'
 apply from: 'gradle/nullability-conventions.gradle'
 
+// Commercial build support: applied only when the COMMERCIAL repository variable is
+// propagated into the environment (i.e. in the spring-pulsar-commercial repository).
+// Inert in OSS builds where COMMERCIAL is unset.
+if (System.getenv('COMMERCIAL')) {
```
artembilan: Yeah... This feels a bit fragile.
I my understand why you cannot rely on the `ARTIFACTORY_USERNAME` like we do in other projects but then this `COMMERCIAL` has to be propagated in more places to the ENV.
SNAPSHOT build, PR build, release train `build` and `test`.

Or... the logic in those workflows has to be revisited to not use `ARTIFACTORY_USERNAME`...

artembilan (changes requested):
Thanks for doing this, Soby!

But as we have discussed many times: it is not that easy since the project diverges from our experience in other projects.

Let's hope we can nail it anyway!

.github/workflows/post-release.yml
sobychacko: maybe we should eventually drop the `perfom_release` from ci.yml.

.github/workflows/post-release.yml
sobychacko: Looks good to me from what i can see.

.github/workflows/post-release.yml
artembilan: I think that should be a separate task.
For now we need to assume that release-related code in the `ci.yml` is dead because we are not going to release manually anymore.

build.gradle
sobychacko: I tried to rework it so that it is now gated on the `ARTIFACTORY_USERNAME` condition.

